PR #17, opened 2020-12-16
[WIP] consistent API

rus-alex:
Node provides only committed data for external API.
Committed data is a consistent and already flushed on disc blockchain info and notifications. It prevent possibility to read uncompleted data (from the middle of consensus process) and to get notifications twice (after node restart).

WIP: temporary depends on lachesis-base fork until https://github.com/Fantom-foundation/lachesis-base/pull/13


rus-alex:
Rejected as far as speed of txs confirming is more important than consistency.

PR #44, opened 2021-02-18
No discovery v5

rus-alex:
Disabled discovery v5 by default according to [recommendations](https://github.com/ethereum/go-ethereum/blob/v1.9.25/p2p/discv5/README) and memory leaks research results.


rus-alex:
discV5 is nice enough

PR #85, opened 2021-03-31
Event export API method

rus-alex:
API method 'debug_exportEvents' doubles command `./opera export ...` functionality, but allows to do it without node stopping.
Helps to make more simple and reliable snapshots server, for example.


rus-alex:
Outdated

PR #171, opened 2021-09-27
Some refactoring

rus-alex:
Note: changes include and depend on #163 .
The main points are:
 - func СheckEvm() is moved from cmd/opera/launcher to gossip/evmstore package;
 - genesis file is needed only for chain data (datadir) initialization;
 - genesis store is dropped after chain data (datadir) initialization;
 - BlockEpochState history stores data on the each epoch begin;
 - a bit more readable epoch sealer;


uprendis:
This PR was an experimental attempt to implement the manual trustful "trustpoints" functionality. The PR gets closed in favor of the LLR approach but we will reuse some of the refactoring commits in future PRs

PR #186, opened 2021-11-12
fix of gossip.Store.recoverBlockState()

rus-alex:
Update to migrate 'BlockEpochStateHistory' table also.


uprendis:
The reason why there were no migration for BlockEpochStateHistory is because this table didn't exist in previous released versions. Nevertheless, we can add it just-in-case

PR #200, opened 2021-12-20
LLR fixes

uprendis:
- For LLR blocks voting, each epoch has a separate voting tally to prevent single validator from using an old epoch N times to decode the voting
- For LLR epochs/blocks voting tally, vote from the same validator cannot be counted twice. Use bitmap to store indexes of voted validators
- Numerous fixes/improvements for Misbehavior proofs (MPs)
- LLR emitting is optimized so that poor validators emit LLR votes only when top validators failed to emit enough votes
- gossip tests use consensus engine instead of a mocked engine
- refactor LLR/MPs checkers
- event transactions are serialized with RLP starting from event version 1
- add support for multiple event emitters at once
- add forward compatibility for NetForkID. This field is intended to prevent slashing future non-upgraded validators in a case if they emit wrong LLR votes


cmd/opera/launcher/fake.go
```diff
@@ -19,19 +19,19 @@ var FakeNetFlag = cli.StringFlag{
 }
 
 func getFakeValidatorKey(ctx *cli.Context) *ecdsa.PrivateKey {
-	num, _, err := parseFakeGen(ctx.GlobalString(FakeNetFlag.Name))
+	id, _, err := parseFakeGen(ctx.GlobalString(FakeNetFlag.Name))
 	if err != nil {
```
cyberbono3, line 23: I suggest to combine both cases into one:  ` if err != nil ||  id == 0 {  return nil }`

cmd/opera/launcher/validator.go
```diff
@@ -33,38 +33,28 @@ var validatorPasswordFlag = cli.StringFlag{
 // command line flags or from the keystore if CLI indexed.
 func setValidator(ctx *cli.Context, cfg *emitter.Config) error {
```
cyberbono3, line 34: should not we call error check logic at the beginning of `setValidator`  or ? 
```
func setValidator(ctx *cli.Context, cfg *emitter.Config) error {
      	if cfg.Validator.ID != 0 && cfg.Validator.PubKey.Empty() {
		      return errors.New("validator public key is not set")
	} 
    ....
   
}

File: cmd/opera/launcher/validator.go [hunk 3]
```diff
@@ -33,38 +33,28 @@ var validatorPasswordFlag = cli.StringFlag{
...
 	// Extract the current validator address, new flag overriding legacy one
-	var validatorID idx.ValidatorID
-	var validatorPubkey validatorpk.PubKey
 	var err error
 	if ctx.GlobalIsSet(FakeNetFlag.Name) {
```
cyberbono3, line 37: Just curious what will happen if both `ctx.GlobalIsSet(FakeNetFlag.Name)` and `ctx.GlobalIsSet(validatorIDFlag.Name)` evaluate to `true` ? First case one updates `cfg.Validator.ID, num, err = parseFakeGen(ctx.GlobalString(FakeNetFlag.Name))`, the second one would override  `cfg.Validator.ID`  with `idx.ValidatorID(ctx.GlobalInt(validatorIDFlag.Name))`. Just wanna make sure there is no conflict here.

cmd/opera/launcher/validator.go
```diff
@@ -33,38 +33,28 @@ var validatorPasswordFlag = cli.StringFlag{
 // command line flags or from the keystore if CLI indexed.
 func setValidator(ctx *cli.Context, cfg *emitter.Config) error {
 	// Extract the current validator address, new flag overriding legacy one
-	var validatorID idx.ValidatorID
-	var validatorPubkey validatorpk.PubKey
 	var err error
```
cyberbono3, line 36: I think we can remove err variable on L36 and use `err` variable inside if blocks.

gossip/c_block_callbacks_test.go
```diff
@@ -16,47 +17,51 @@ func TestConsensusCallback(t *testing.T) {
...
 
-	for n := uint64(0); n < blockCount; n++ {
+	for n := uint64(0); n < rounds; n++ {
 		// transfers
 		txs := make([]*types.Transaction, accounts)
 		for i := range txs {
 			from := (i)%accounts + 1
-			to := (i+1)%accounts + 1
-			txs[i] = env.Transfer(from, to, utils.ToFtm(100))
+			to := 0 + 1
```
cyberbono3, line 40: what you mean by writing  `0+1` instead of just `1`?

gossip/dummy_tx_pool.go
```diff
@@ -99,3 +143,27 @@ func (p *dummyTxPool) SampleHashes(max int) []common.Hash {
...
+	if len(p.pool) != 0 {
+		p.pool = p.pool[:0]
+	}
+}
+
+func (p *dummyTxPool) Delete(needle common.Hash) {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	notErased := make([]*types.Transaction, 0, len(p.pool)-1)
+	for _, tx := range p.pool {
```
cyberbono3, line 163: can we set `notErased[i] = tx`  instead of `notErased = append(notErased, tx)` as capacity of `notErased` is known?
```
notErased := make([]*types.Transaction, 0, len(p.pool)-1)
for i, tx := range p.pool {
	if tx.Hash() != needle {
		notErased[i] = tx
	}
}
```

gossip/store_decided_state.go
```diff
@@ -17,30 +17,49 @@ type BlockEpochState struct {
...
+	s.cache.BlockEpochStateHistory.Add(epoch, bes, nominalSize)
 }
 
 func (s *Store) GetHistoryBlockEpochState(epoch idx.Epoch) (*iblockproc.BlockState, *iblockproc.EpochState) {
-	// check current BlockEpochState as a cache
-	if v := s.cache.BlockEpochState.Load(); v != nil {
+	// Get HistoryBlockEpochState from LRU cache first.
+	if v, ok := s.cache.BlockEpochStateHistory.Get(epoch); ok {
 		bes := v.(*BlockEpochState)
 		if bes.EpochState.Epoch == epoch {
```
cyberbono3, line 34: I suggest to simplify the code here
```
	if v, ok := s.cache.BlockEpochStateHistory.Get(epoch); ok {
		bes := v.(*BlockEpochState)
		if bes.EpochState.Epoch == epoch {
			return bes.BlockState,bes.EpochState
		}
```

inter/event_serializer.go
```diff
@@ -17,8 +17,11 @@ import (
 var (
 	ErrSerMalformedEvent = errors.New("serialization of malformed event")
 	ErrTooLowEpoch       = errors.New("serialization of events with epoch<256 and version=0 is unsupported")
+	ErrUnknownVersion    = errors.New("unknown serialization version")
```
cyberbono3, line 20: why you made `ErrUnknownVersion` public ?  It is used only within `inter` package here:
```
	if version > MaxSerializationVersion {
		return ErrUnknownVersion
	}
```

cyberbono3: ditto  `ErrSerMalformedEvent` and `ErrTooLowEpoch`

cyberbono3 (approved):
Overall, LGTM. Left some comments.

cmd/opera/launcher/validator.go
uprendis: `setValidator` method updates `cfg.Validator`. `cfg.Validator` may be (and normally is) empty before calling `setValidator`
The goal of the check is to enforce that "if validator ID is set, then validator pubkey must be set too"

gossip/c_block_callbacks_test.go
uprendis: `from` and `to` store validator IDs (normally typed as idx.ValidatorID), `i` and `accounts` store indexes of validators (normally typed as idx.Validator). Validator ID can never be 0. Normally (it depends on the SFC contract implementation) validator IDs start from 1 and increment after each new validator. In a general case, you cannot take validator index, add 1 and expect it to be his validator ID, because:
- Validators are sorted by stake first, so even a younger validator may have a smaller index than an older validator
- Some validators get deactivated, so validator IDs are sparse in a general case. E.g. we can have validator IDs `{1, 4, 9}`

In the test, we can just add 1 because all validators have the same stake and never get deactivated. Normally, one would need to call `from := env.store.GetValidators().GetID(idx.Validator((i)%accounts + 1))`. We write `0+1` instead of `1` to highlight that it's ID of first validator

In this test, the typing is not very strict, as both `i` and `accounts` are typed as int. It may make sense to update the typing to make it more obvious what variables store

gossip/store_decided_state.go
uprendis: By doing this
```			
bs := *bes.BlockState
es := *bes.EpochState
return &bs, &es
```
We want to ensure that returned values have different address, so that caller couldn't accidentally modify the value in cache. Now when I think about it, we should actually call
```			
bs := bes.BlockState.Copy()
es := bes.EpochState.Copy()
return &bs, &es
```
So that .Copy would also make a copy of underlying arrays and pointers

cmd/opera/launcher/validator.go
uprendis: If both `--fakenet` and `--validator.id` are set, then `--validator.id` will overrule fakenet's validator ID. The same with `--validator.pubkey`
There are situations when you need to use both `--fakenet` and `--validator.id`/`--validator.pubkey`, e.g.:
- when you staked a new validator on fakenet and need to set up a validator for it. You would need to write something like `--fakenet 0/3 --validator.id 4 --validator.pubkey 0xsomehex`. Yet, I don't think there are any situations when you have to use `--fakenet` with non-zero first argument (e.g. `--fakenet 1/3`) and --validator.id. So we may restrict it, e.g.:
```
	if ctx.GlobalIsSet(validatorIDFlag.Name) && ctx.GlobalIsSet(FakeNetFlag.Name) && cfg.Validator.ID != 0 {
		return errors.New("specified validator ID with both --fakenet and --validator.id")
	}
	if ctx.GlobalIsSet(validatorIDFlag.Name) {
```
Do you think we should add such restriction?

cmd/opera/launcher/validator.go
uprendis: We would need 2 lines instead of 1 in this case (there's 2 if blocks which use `err`). Do you think we should do it?

inter/event_serializer.go
uprendis: The most `ErrX` variables in this repo are public but never used outside their package. If we were to refactor it, we may need a separate PR

cmd/opera/launcher/validator.go
cyberbono3: We can do like that:
```
	// var err error
	if ctx.GlobalIsSet(FakeNetFlag.Name) {
		var num idx.Validator
		id, num, err := parseFakeGen(ctx.GlobalString(FakeNetFlag.Name))
		if err != nil {
			return err
		}
		cfg.Validator.ID = id 
		validators := makegenesis.GetFakeValidators(num)
		cfg.Validator.PubKey = validators.Map()[cfg.Validator.ID].PubKey
	} 

	if ctx.GlobalIsSet(validatorIDFlag.Name) {
		cfg.Validator.ID = idx.ValidatorID(ctx.GlobalInt(validatorIDFlag.Name))
	}
	if ctx.GlobalIsSet(validatorPubkeyFlag.Name) {
		pk, err := validatorpk.FromString(ctx.GlobalString(validatorPubkeyFlag.Name))
		if err != nil {
			return err
		}
		cfg.Validator.PubKey = pk
	}
```

cmd/opera/launcher/validator.go
cyberbono3: yeah we have to restrict this unexpected behavior `--fakenet 0/3 --validator.id 4 --validator.pubkey 0xsomehex` if there are no situations where we can set all three flags at once. I can open a separate PR for it. Does it sound good to you?

cmd/opera/launcher/validator.go
uprendis: Yeah, looking good

cmd/opera/launcher/validator.go
uprendis: You misunderstood - there are some cases when using all 3 flags will make sense, and other cases when it has no sense
In your example, it may make sense. User may use `--fakenet 0/3` just to specify the network, and then `--validator.id 4 --validator.pubkey 0xsomehex` to specify validator. Normally, when you're working in fakenet, there's a shortcut for that - you can just use `--fakenet 1/3` if you also want to run a validator. However, fakenet is an unrestricted network, and you may create new validators at will. This shortcut (`--fakenet 1/3` ) works only for genesis fakenet validators, for a newly created validator user has to use the `--validator.id 4 --validator.pubkey 0xsomehex`

uprendis:
```
	if ctx.GlobalIsSet(validatorIDFlag.Name) && ctx.GlobalIsSet(FakeNetFlag.Name) && cfg.Validator.ID != 0 {
		return errors.New("specified validator ID with both --fakenet and --validator.id")
	}
	if ctx.GlobalIsSet(validatorIDFlag.Name) {
```
In other words, if user specified `fakenet` with some non-zero validator, and then proceeded with overriding this validator with `--validator.id`, then it doesn't have a lot of sense (at least to me). It would be clearer if user specified `fakenet` with zero validator in this case

cyberbono3: Thanks for clarification. Yeah, I lack some context here. 
I assume, this case makes no sense. That's why you yield an error here.
```
if ctx.GlobalIsSet(validatorIDFlag.Name) && ctx.GlobalIsSet(FakeNetFlag.Name) && cfg.Validator.ID != 0 {
	return errors.New("specified validator ID with both --fakenet and --validator.id")
}
```
You mentioned there are more cases where it does not make sense. Would you like to add them as well?

uprendis: I believe this `if block` addresses all "silly" cases

PR #206, opened 2021-12-22
Fix switching epoch after snapsync

uprendis:
- EVM snapshots generation is done in background to prevent prolonged blocking during switching epoch by snapsync. Add periodic DBs flushing due to the new background DB writes done by EVM snapshots generation
- wait until block processing has finished (edge case) before switching epoch
- switch epoch in engine's storage
- force DB committing after switching epoch


gossip/c_event_callbacks.go
```diff
@@ -143,13 +143,21 @@ func (s *Service) SwitchEpochTo(newEpoch idx.Epoch) error {
...
 	defer s.engineMu.Unlock()
+	s.blockProcWg.Wait()
 	if newEpoch == s.store.GetEpoch() {
 		return errSameEpoch
 	}
 	s.store.SetBlockEpochState(*bs, *es)
 	s.switchEpochTo(newEpoch)
-	_ = s.store.EvmSnapshotAt(common.Hash(bs.FinalizedStateRoot))
-	return nil
+	err := s.engine.Reset(newEpoch, es.Validators)
```
cyberbono3, line 152: When I checked out the branch, it yields the error  `s.engine.Reset undefined`

hadv: @cyberbono3 FYI: I get the branch without the error

gossip/c_event_callbacks.go
```diff
@@ -143,13 +143,21 @@ func (s *Service) SwitchEpochTo(newEpoch idx.Epoch) error {
...
+	if err != nil {
+		return err
+	}
+	err = s.store.EvmSnapshotAt(common.Hash(bs.FinalizedStateRoot))
```
hadv, line 156: nitpick: it might not the issue but it sound more logically if we move the snapshot generating before switching to new epoch

gossip/evmstore/store.go
```diff
@@ -103,14 +104,17 @@ func (s *Store) initCache() {
 	s.cache.EvmBlocks = s.makeCache(s.cfg.Cache.EvmBlocksSize, s.cfg.Cache.EvmBlocksNum)
 }
 
-func (s *Store) CreateEvmSnapshot(root common.Hash) (err error) {
+func (s *Store) GenerateEvmSnapshot(root common.Hash, rebuild, async bool) (err error) {
```
hadv, line 107: just a coding convention, it's clearer if we specific datatype for each parameter separately. 
By the way, I'm OK with the current one if everyone feel good with it :) 
 
```suggestion
func (s *Store) GenerateEvmSnapshot(root common.Hash, rebuild bool, async bool) (err error) {
```

gossip/evmstore/store.go
```diff
@@ -119,7 +123,6 @@ func (s *Store) RebuildEvmSnapshot(root common.Hash) {
 	if s.Snaps == nil {
```
hadv, line 123: Question: if we check like that then when the snapshot rebuild can be processed?

uprendis: If s.Snaps is nil, then it'll have to call GenerateEvmSnapshot with `rebuild`=true and `async`=true. It should have almost the same effect as calling s.Snaps.Rebuild

hadv: Yes. I mean if the s.Snaps is nil then it never generate a new snapshot with new hash root?

uprendis: If the s.Snaps is not nil, then it'll call s.Snaps.Rebuild, which will erase existing data (if any) and generate new snapshot in background with new hash root. We'll proceed straight to full syncing, each new processed block will update generation of the snapshot. Eventually, snapshot should be generated during full syncing

hadv (approved):

PR #207, opened 2021-12-23
Update setValidator

cyberbono3:
- add an error for all "silly cases"
- remove `error` variable in `setValidator`

### DIFF
--- a/cmd/opera/launcher/validator.go
+++ b/cmd/opera/launcher/validator.go
@@ -32,15 +32,17 @@ var validatorPasswordFlag = cli.StringFlag{
 // setValidatorID retrieves the validator ID either from the directly specified
 // command line flags or from the keystore if CLI indexed.
 func setValidator(ctx *cli.Context, cfg *emitter.Config) error {
-	if ctx.GlobalIsSet(validatorIDFlag.Name) && ctx.GlobalIsSet(FakeNetFlag.Name) && cfg.Validator.ID != 0 {
-		return errors.New("specified validator ID with both --fakenet and --validator.id")
-	}
 	// Extract the current validator address, new flag overriding legacy one
 	if ctx.GlobalIsSet(FakeNetFlag.Name) {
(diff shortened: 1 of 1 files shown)


cmd/opera/launcher/validator.go
```diff
@@ -32,25 +32,29 @@ var validatorPasswordFlag = cli.StringFlag{
 // setValidatorID retrieves the validator ID either from the directly specified
 // command line flags or from the keystore if CLI indexed.
 func setValidator(ctx *cli.Context, cfg *emitter.Config) error {
+	if ctx.GlobalIsSet(validatorIDFlag.Name) && ctx.GlobalIsSet(FakeNetFlag.Name) && cfg.Validator.ID != 0 {
```
uprendis, line 35: This check has to be after `if ctx.GlobalIsSet(FakeNetFlag.Name) {`, or better inside that `if block`

cmd/opera/launcher/validator.go
```diff
@@ -32,25 +32,29 @@ var validatorPasswordFlag = cli.StringFlag{
...
 	// Extract the current validator address, new flag overriding legacy one
-	var err error
 	if ctx.GlobalIsSet(FakeNetFlag.Name) {
-		var num idx.Validator
-		cfg.Validator.ID, num, err = parseFakeGen(ctx.GlobalString(FakeNetFlag.Name))
+		id, num, err := parseFakeGen(ctx.GlobalString(FakeNetFlag.Name))
 		if err != nil {
 			return err
 		}
+		cfg.Validator.ID = id
```
uprendis, line 44: We can move the check `if ctx.GlobalIsSet(validatorIDFlag.Name) && ctx.GlobalIsSet(FakeNetFlag.Name) && cfg.Validator.ID != 0 {` here. We can simplify it to `if ctx.GlobalIsSet(validatorIDFlag.Name) && id != 0 {`

PR #208, opened 2021-12-23
Minor code improvements - WIP

cyberbono3:


gossip/emitter/control.go
```diff
@@ -62,6 +76,7 @@ func (em *Emitter) isAllowedToEmit(e inter.EventI, eTxs bool, metric ancestor.Me
 	adjustedPassedIdleTime := time.Duration(ancestor.Metric(passedTimeIdle/piecefunc.DecimalUnit) * metric)
 	passedBlocks := em.world.GetLatestBlockIndex() - em.prevEmittedAtBlock
 	// Forbid emitting if not enough power and power is decreasing
+	// Question: what is the intetion to declare code blocks with {...threshold = } to keep variables only within code block?
```
uprendis, line 79: Yeah, just to separate the variables. Also visually separate the code

gossip/emitter/emitter.go
```diff
@@ -127,13 +128,11 @@ func (em *Emitter) init() {
 
 // Start starts event emission.
 func (em *Emitter) Start() {
-	if em.config.Validator.ID == 0 {
+	if em.config.Validator.ID == 0 || em.done != nil  {
```
uprendis, line 131: Will need to update the comment to "short circuit if not a validator or already started"

gossip/emitter/sync.go
```diff
@@ -51,8 +51,7 @@ func (em *Emitter) currentSyncStatus() doublesign.SyncStatus {
 	if em.world.IsSynced() {
 		s.P2PSynced = em.syncStatus.p2pSynced
 	}
-	prevEmitted := em.readLastEmittedEventID()
-	if prevEmitted != nil && (em.world.GetEvent(*prevEmitted) == nil && em.epoch <= prevEmitted.Epoch()) {
+	if prevEmitted := em.readLastEmittedEventID(); prevEmitted != nil && (em.world.GetEvent(*prevEmitted) == nil && em.epoch <= prevEmitted.Epoch()) {
```
uprendis, line 54: Hmm, this one became too long

gossip/emitter/txs.go
```diff
@@ -157,11 +156,10 @@ func (em *Emitter) addTxs(e *inter.MutableEventPayload, sorted *types.Transactio
 	}
 
 	// sort transactions by price and nonce
-	rules := em.world.GetRules()
 	for tx := sorted.Peek(); tx != nil; tx = sorted.Peek() {
 		sender, _ := types.Sender(em.world.TxSigner, tx)
 		// check transaction epoch rules
-		if epochcheck.CheckTxs(types.Transactions{tx}, rules) != nil {
+		if epochcheck.CheckTxs(types.Transactions{tx}, em.world.GetRules()) != nil {
```
uprendis, line 162: nitpick: it might be less optimal to call the function in loop

gossip/c_event_callbacks.go
```diff
@@ -196,12 +196,15 @@ func (s *Service) processEvent(e *inter.EventPayload) error {
 	}
 
 	newEpoch := s.store.GetEpoch()
+	newEpochNotEqualOld := newEpoch != oldEpoch
```
uprendis, line 199: You can name "sealed" or "epochSealed". "Epoch is sealed" means that it had ended and a next epoch has started

gossip/emitter/control.go
```diff
@@ -29,6 +29,20 @@ func updMetric(median, cur, upd idx.Event, validatorIdx idx.Validator, validator
...
+
+	 }
+
+	 return metrics
+
+	what if seq = 0, and metric = 0( i dont know if it possible) so this would satisfy both if conditions 
+	and metric will be increased twice, is that expected behavior?
+
+	*/
+	// what will happen if e.g. seq = 1, metric = 0
```
uprendis, line 45: We'll return 0.3

gossip/emitter/control.go
```diff
@@ -29,6 +29,20 @@ func updMetric(median, cur, upd idx.Event, validatorIdx idx.Validator, validator
 func eventMetric(orig ancestor.Metric, seq idx.Event) ancestor.Metric {
 	metric := ancestor.Metric(eventMetricF(uint64(orig)))
 	// kick start metric in a beginning of epoch, when there's nothing to observe yet
+	/* Question: are 2 if conditions below mutual exclusive or not? if they are mutual exclusive, switch statement can be a good option
```
uprendis, line 32: Not exclusive

gossip/emitter/emitter.go
```diff
@@ -232,15 +233,12 @@ func (em *Emitter) getSortedTxs() *types.TransactionsByPriceAndNonce {
...
 
 func (em *Emitter) EmitEvent() (*inter.EventPayload, error) {
-	if em.config.Validator.ID == 0 {
+	if em.config.Validator.ID == 0 || em.world.IsBusy() {
 		// short circuit if not a validator
 		return nil, nil
 	}
 	sortedTxs := em.getSortedTxs()
 
-	if em.world.IsBusy() {
```
uprendis, line 241: It may make sense to leave the check closer to `em.world.Lock()` to reduce the chance of locking the mutex when there's an active block processing. The reason why we bother with `IsBusy` check is because `em.world.Lock` will lock engine mutex and then wait until block processing is complete (if any). That will make engine mutex unavailable while we're waiting for completion of block processing (in a case if it's running)

gossip/emitter/prev_action_files.go
```diff
@@ -82,24 +71,35 @@ func (em *Emitter) writeLastEmittedEpochVote(e idx.Epoch) {
...
+}
+
+// should we put logger on *Emitter?
+func readFile(file *os.File, bufLen uint8, errMsg, path string) []byte {
+	if file == nil {
+		return nil
+	}
+
+	buf := make([]byte, bufLen)
+	if _, err := file.ReadAt(buf, 0); err != nil {
```
uprendis, line 96: Technically we had/have a bug here, because ReadAt is not obliged to read the entire buffer. Practically it works only because buffer is very small

uprendis: We may use `Seek` and then call `ioread.ReadAll` to read entire buffer. Or, we can wrap it with io.NewSectionReader (so that we wouldn't need to call `Seek`) and then call `ioread.ReadAll`

uprendis: Oh sorry, I'm wrong:
```
// When ReadAt returns n < len(p), it returns a non-nil error
// explaining why more bytes were not returned. In this respect,
// ReadAt is stricter than Read.
```

gossip/emitter/sync.go
cyberbono3: Does multiline if statement sound good to you?  It does not make sense to declare `prevEmitted` as local in `currentSyncStatus` since it is used only inside if statement.
```
if prevEmitted := em.readLastEmittedEventID(); prevEmitted != nil &&
		em.world.GetEvent(*prevEmitted) == nil &&
		em.epoch <= prevEmitted.Epoch() {
		s.P2PSynced = time.Time{}
	}
```

uprendis: ```It does not make sense to declare prevEmitted as local```
Makes sense! The last variant looks good

gossip/emitter/emitter.go
cyberbono3: I get your point. Reverting my changes.

cyberbono3:
@uprendis  I have addressed your comments. I will continue working on it as soon as I have compete tasks with higher priority. Let us leave in draft for now.

PR #210, opened 2021-12-29
clean up and commit state trie for snapsync

hadv:
Previously, by default opera only support `archive` gcmode, this PR will enable the `--gcmode` for opera node with two new options after introducing the `snapsync` 

1. `--gcmode=light`: prune trie state in memory cache
2. `--gcmode=full`: prune trie state both in memory cache and in persistent db

And add new command to run compact db on non-running node.


gossip/evmstore/store.go
```diff
@@ -126,6 +126,31 @@ func (s *Store) RebuildEvmSnapshot(root common.Hash) {
 	s.Snaps.Rebuild(root)
 }
 
+// Commit changes.
+func (s *Store) CleanCommit(block iblockproc.BlockState) error {
```
uprendis, line 130: We may change it so that `CleanCommit` would get executed inside `evmstore.Store.Commit` if `flush=true`. It seems that we should either have 2 functions (`CleanCommit` and `Commit`) but without the `flush` argument, or only `Commit` function with the `flush` argument

gossip/evmstore/store.go
```diff
@@ -126,6 +126,31 @@ func (s *Store) RebuildEvmSnapshot(root common.Hash) {
...
+	// Don't need to reference the current state root
+	// due to it already be referenced on `Commit()` function
+	triedb := s.EvmState.TrieDB()
+	stateRoot := common.Hash(block.FinalizedStateRoot)
+	current := uint64(block.LastBlock.Idx)
+	// Garbage collect all below the current block
+	for !s.triegc.Empty() {
+		root, number := s.triegc.Pop()
```
uprendis, line 138: If we dereference all roots but the latest one, then it conflicts a little with `evmstore.Store.Flush`. `evmstore.Store.Flush` tries to commit `head`, `head-1`, `head-31` roots. All of them but `head` will already be dereferenced, so we can simplify it and commit only the `head` inside `Flush`. `Flush` also won't need the `getBlock` argument anymore

cmd/opera/launcher/launcher.go
```diff
@@ -116,6 +116,7 @@ func initFlags() {
 		validatorPubkeyFlag,
 		validatorPasswordFlag,
 		SyncModeFlag,
+		utils.GCModeFlag,
```
uprendis, line 119: Let's make a copy of `utils.GCModeFlag` but with default value = `archive`. It's because some existing nodes rely on the absence of GC, and will be overwhelmed if we implicitly enable GC

gossip/evmstore/store.go
uprendis: Also, if we dereference all roots bit the latest one, it will make them inaccessible by API calls (if I understand correctly). Then, we don't respect the `TriesInMemory` constant. As far as I understand, we should either make `TriesInMemory`=1 or not dereference roots inside `CleanCommit`

cmd/opera/launcher/launcher.go
hadv: okay, I got it

gossip/evmstore/store.go
hadv: @uprendis Yes, I think in our case, we only need to commit only the `head` inside `Flush` due to go-opera don't need to re-org if I understand correctly. We might change `TriesInMemory`=1. I will consolidate all the related things and commit the code afterward. Thanks!

gossip/evmstore/store.go
hadv: But with the current code if `flush=true` mean we does not clean any things but commit everything to the db like archive mode

gossip/evmstore/store.go
hadv: @uprendis by the way, on the 2nd thought, if we periodic call the commit every 10 seconds with `period: 10 * time.Millisecond` in the `PeriodicFlusher` then I think we don't have any effective way to pruning the state on mem-cache. Or I mis-understand smth on here.

uprendis: We don't prune every 10 ms, we check if a commit is needed every 10ms. flushable.Store tracks the size of in-memory diff. We sum those sizes for every DB and compare it to a configured threshold. Practically we commit every ~30-100 blocks on mainnet.
But yes, it does mean that geth's pruning strategy may be not very effective for us, because it's based only on avoiding writing MPT data (unless node stops, or `TrieTimeLimit` passed since prev flush, or `Cap` is called and dirties map is too large). All we can archive with geth's approach is that we will flush every ~30-100th root rather than every root (which is still better than nothing, but not ideal in terms of data pruning). To make a more greedy pruning, we would need to not only avoid writing MPT data, but also erase previously written MPT data if it's not referenced anymore by latest root(s)

hadv: Yes, but it seems that the `s.dbs.NotFlushedSizeEst()` quickly return a big number (seems accumulating) then the commit happen very frequency via `mayCommit()` method call

hadv: @uprendis I debug log and found that with current **snapsync**, the latest block state is always at block number 1

```
// GetBlockState retrieves the latest block state
func (s *Store) GetBlockState() iblockproc.BlockState {
	return *s.getBlockEpochState().BlockState
}
```

So the below **commitEVM** invoke has no meaning for state commit for the current code on **develop** branch

```
	// TODO: prune old MPTs in beginnings of committed sections
	if !s.store.cfg.EVM.Cache.TrieDirtyDisabled {
		s.store.commitEVM(true)
	}
```

**_And I think we also cannot make the trie state pruning by using the current triegc due to in that way, we need to record every block's state root into the triegc queue then doing the pruning later._** but with snapsync, we didn't record the state root into the triegc.

uprendis:
```
I debug log and found that with current snapsync, the latest block state is always at block number 1
```
Yeah, that's right. While snapsync hasn't finished, current block number won't move. After snapsync, we'll jump from an epoch which we had before the snapsync, to the new epoch
Yes, Service.commit during snapsync will only flush the content of flushable.Store's. s.store.commitEVM will practically do nothing

uprendis:
```
And further more I think we also cannot make the trie state pruning by using the current triegc due to in that way, we need to record every block's state root into the triegc queue then doing the pruning later. but with snapsync, we didn't record the state root into the triegc.
```
Hmm, I don't quite understand. After snapsync, the MPTs will already be pruned, because we downloaded only latest roots. After snapsync, triegc should work (not effectively as it'll commit every ~30-100th root)?

hadv: > 
> Hmm, I don't quite understand. After snapsync, the MPTs will already be pruned, because we downloaded only latest roots. After snapsync, triegc should work (not effectively as it'll commit every ~30-100th root)?

Yes, I were thinking about when the snapsync in-progress so I though like that. After snapsync it be work like you said.

hadv:
ouch, node with full sync sudden stuck at snapshot generating forever 

```
INFO [01-07|11:12:45.791] Generating state snapshot                root=caa5dd..1ebe32 at=d1901f..f8aa13 accounts=42,317,571 slots=66,216,415 storage=5.78GiB elapsed=2m0.023s   eta=5m15.538s
INFO [01-07|11:12:45.830] Looking for peers                        peercount=0 tried=1086 static=0
INFO [01-07|11:12:53.792] Generating state snapshot                root=caa5dd..1ebe32 at=d2d3ce..9c1d6d accounts=42,572,856 slots=66,491,415 storage=5.80GiB elapsed=2m8.023s   eta=5m5.535s
INFO [01-07|11:13:01.795] Generating state snapshot                root=caa5dd..1ebe32 in=d40fab..ac5a31 at=762de4..c9136b accounts=42,822,251 slots=66,809,668 storage=5.83GiB elapsed=2m16.026s  eta=4m56.444s
INFO [01-07|11:13:09.796] Generating state snapshot                root=caa5dd..1ebe32 at=d53788..c39702 accounts=43,055,928 slots=67,191,675 storage=5.87GiB elapsed=2m24.027s  eta=4m49.055s
INFO [01-07|11:13:17.796] Generating state snapshot                root=caa5dd..1ebe32 at=d677b9..435f5b accounts=43,309,251 slots=67,500,115 storage=5.90GiB elapsed=2m32.027s  eta=4m39.776s
INFO [01-07|11:13:19.253] New LLR summary                          last_epoch=16079 last_block=907951 new_evs=2 new_ers=1 new_bvs=12 new_brs=0 age=3m12.998s
INFO [01-07|11:13:19.253] New DAG summary                          new=2 last_id=16066:1:994f21 age=29m8.341s  t="235.778µs"
WARN [01-07|11:13:19.253] Incoming event rejected                  event=16066:1:994f21 creator=3 err="wrong event epoch hash"
INFO [01-07|11:13:25.796] Generating state snapshot                root=caa5dd..1ebe32 at=d7bff6..28eec7 accounts=43,568,631 slots=67,702,075 storage=5.92GiB elapsed=2m40.027s  eta=4m30.063s
INFO [01-07|11:13:33.796] Generating state snapshot                root=caa5dd..1ebe32 at=d8ee44..1e5bcb accounts=43,808,056 slots=68,099,395 storage=5.95GiB elapsed=2m48.028s  eta=4m22.261s
INFO [01-07|11:13:41.801] Generating state snapshot                root=caa5dd..1ebe32 in=da3749..85b110 at=b8f3a9..a1f4c9 accounts=44,066,850 slots=68,375,004 storage=5.98GiB elapsed=2m56.032s  eta=4m12.739s
INFO [01-07|11:13:49.801] Generating state snapshot                root=caa5dd..1ebe32 at=db77d3..5c8763 accounts=44,319,573 slots=68,590,215 storage=6.00GiB elapsed=3m4.032s   eta=4m3.866s
INFO [01-07|11:13:57.801] Generating state snapshot                root=caa5dd..1ebe32 at=dcca00..146704 accounts=44,587,576 slots=68,777,655 storage=6.02GiB elapsed=3m12.032s  eta=3m54.05s
WARN [01-07|11:14:02.632] Served parity_netPeers                   conn=217.12.208.104:56528 reqid=1 t="17.043µs"  err="the method parity_netPeers does not exist/is not available"
INFO [01-07|11:14:05.801] Generating state snapshot                root=caa5dd..1ebe32 at=ddc2cf..dbf596 accounts=44,783,936 slots=69,310,495 storage=6.07GiB elapsed=3m20.032s  eta=3m49.355s
INFO [01-07|11:14:13.801] Generating state snapshot                root=caa5dd..1ebe32 at=dee52a..0e9147 accounts=45,013,176 slots=69,803,295 storage=6.11GiB elapsed=3m28.032s  eta=3m42.187s
```

hadv:
with clean up code (a.k.a `--gcmode=full`) the disk i/o is better than archive gcmode 

**FULL**: _average ~40MB/s_
Screen Shot 2022-01-10 at 18 23 50

**ARCHIVE**: _average +50MB/s_

Screen Shot 2022-01-10 at 18 24 08

hadv:
After **snapsync** (same for both `--gcmode=archive` and `--gcmode=full`), `--gcmode=full` still save a little bit disk i/o and total disk space than `--gcmode=archive`. 

**_Below are metrics on the last 6 hours._**

**FULL**: disk i/o ~45MB/s, total disk space: ~55GB
Screen Shot 2022-01-24 at 10 04 51

**ARCHIVE**: disk i/o ~ 65MB/s, total disk space: ~58GB
Screen Shot 2022-01-24 at 10 05 15

hadv:
We can see the regular up/down of leveldb persistent size.  I think that's the effective of the pruning old state trie on disk
Screen Shot 2022-02-24 at 12 03 50

@uprendis Can you have a look on this PR and https://github.com/Fantom-foundation/go-ethereum/pull/29

hadv:
@uprendis I found that currently, we have a cap call on every commitEVM with very low max default size, it's just **22 MB** 

https://github.com/Fantom-foundation/go-opera/blob/10ca669279e26d21133c1d95db26ea41a6479360/gossip/config.go#L244

So the pruning cannot reach the most effective. Of course, we can increase it relatively with the total `--cache` size but I think it's still small. I tried to increase the `MaxNonFlushedSize` default value to **1024 MB** and made a test on mainnet and it bring more effective for pruning. 

if we synced mainnet from the latest snapshot at block `32661715` to block `33800000` then it can save more than **50 GB** disk space compare to the current default one. 

How do you think about it and can we increase the default value to **1024 MB** or at least to **512 MB** like default trie dirty size?

uprendis:
Got you! Let's keep MaxNonFlushedSize the same but introduce a separate config field for `s.evm.Cap(s.cfg.MaxNonFlushedSize/3, s.cfg.MaxNonFlushedSize/4)` instead of `s.cfg.MaxNonFlushedSize/3` and `s.cfg.MaxNonFlushedSize/4`
Could you please check different sizes for this field in a range of 32 MB - 1024 MB?

uprendis:
Upd: we already have the `TrieDirtyLimit` field, we should use it
I think we should erase the `func (s *Store) Cap(max, min int) {` method and just call `s.EvmState.TrieDB().Cap`

hadv:
> Upd: we already have the `TrieDirtyLimit` field, we should use it I think we should erase the `func (s *Store) Cap(max, min int) {` method and just call `s.EvmState.TrieDB().Cap`

okay, I change the code like in this commit https://github.com/hadv/go-opera/commit/48e8d28a3f963c7ef8c9e36cdcd564092786d235, refactoring the `func (s *Store) Cap()` to reuse in difference places.

> Could you please check different sizes for this field in a range of 32 MB - 1024 MB?

Yes, I will check the difference sizes for it. Actually, I already tested some difference values at default, 256 MB and 1024 MB. **256 MB** seems good enough with `--cache=160000` (16 GB) but I tested in same block range but in difference block heights then comparison a bit improperly. I will check difference size at the same block height to avoid comparing apple to orange.

hadv:
@uprendis Basically, the more cache we use, the more dirty trie we can clean. But base on the below metrics table I think that the default value for `TrieDirtyLimit` is `256 MB` is good enough due to if we use much more cache than that the effectiveness of dirty trie cleaning is not significant improvement. If node operator need to be more effective (but not much), they can increase the total cache size.

**old code** (MaxNonFlushedSize)
| cache   |      block      |  dbsize |
|----------|:-------------:|------:|
|22 MB |  32800000 | 378 GB |
|22 MB |  33000000 | 400 GB |

**new code** (TrieDirtyLimit)
- default total cache = 3096 MB

| cache   |      block      |  dbsize |
|----------|:-------------:|------:|
|32 MB |  32800000 | 377 GB |
|32 MB |  33000000 | 398 GB |
|160 MB |  32800000 | 373 GB |
|160 MB |  33000000 | 388 GB |
|256 MB |  32800000 | 372 GB |
|256 MB |  33000000 | 386 GB |

- cache = 16000 MB

| cache   |      block      |  dbsize |
|----------|:-------------:|------:|
|32 MB |  32800000 | 372 GB |
|32 MB |  33000000 | 388 GB |
|64 MB |  32800000 | 372 GB |
|64 MB |  33000000 | 385 GB |
|128 MB |  32800000 | 371 GB |
|128 MB |  33000000 | 385 GB |
|256 MB |  32800000 | 371 GB |
|256 MB |  33000000 | 385 GB |
|512 MB |  32800000 | 371 GB (388875540) |
|512 MB |  33000000 | 385 GB (403147896) |
|1024 MB |  32800000 | 371 GB (388874108) |
|1024 MB |  33000000 | 385 GB (403149576) |

PR #211, opened 2022-01-05
Add receipts root for API calls

uprendis:
- calculate receipts root in API calls if `RPCBlockExt`=`true`

### DIFF
--- a/ethapi/api.go
+++ b/ethapi/api.go
@@ -707,7 +707,7 @@ func (s *PublicBlockChainAPI) calculateExtBlockApi(ctx context.Context, blkNumbe
 		if err != nil {
 			return ext
 		}
-		if len(receipts) != 0 {
+		if receipts.Len() != 0 {
 			ext.receiptsRoot = types.DeriveSha(receipts, new(trie.Trie))
 			ext.bloom = types.CreateBloom(receipts)
 		} else {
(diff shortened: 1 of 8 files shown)


ethapi/api.go
```diff
@@ -694,22 +695,26 @@ func (s *PublicBlockChainAPI) GetHeaderByNumber(ctx context.Context, number rpc.
...
+	var ext extBlockApi
+	if s.b.CalcBlockExtApi() && blkNumber != rpc.EarliestBlockNumber {
 		receipts, err := s.b.GetReceiptsByNumber(ctx, blkNumber)
 		if err != nil {
-			return types.Bloom{}
+			return ext
 		}
-		if receipts != nil {
-			return types.CreateBloom(receipts)
+		if len(receipts) != 0 {
```
hadv, line 710: nitpick: I found that `types.Receipts` already has a wrapper method to get len; so I think it'd better to use it 

```suggestion
		if receipts.Len() != 0 {
```

cyberbono3 (approved):
lgtm

gossip/config.go
```diff
@@ -105,7 +105,7 @@ type (
 
 		ExtRPCEnabled bool
 
-		RPCLogsBloom bool
+		RPCBlockExt bool
```
rus-alex, line 108: Suggestion for the future: move `RPCBlockExt` to the new `ethapi.Config` struct and configure API by constructor `NewPublicBlockChainAPI(Backend, Config)`

rus-alex (approved):

hadv (approved):

PR #216, opened 2022-01-14
Fix snapsync and LLR race conditions

uprendis:
- fix race conditions between snapsync and fullsync
- fix race condition between blocks processing and BR processing

### DIFF
--- a/gossip/apply_genesis.go
+++ b/gossip/apply_genesis.go
@@ -184,7 +184,7 @@ func (s *Store) applyEpoch1Genesis(blockProc BlockProc, g opera.Genesis) (err er
 	s.SetHistoryBlockEpochState(es.Epoch, bs, es)
 	s.SetEpochBlock(blockCtx.Idx+1, es.Epoch)
 	s.SetBlockEpochState(bs, es)
-	s.SetLlrState(LlrState{
+	s.setLlrState(LlrState{
 		LowestEpochToDecide: es.Epoch + 1,
 		LowestEpochToFill:   es.Epoch + 1,
 		LowestBlockToDecide: bs.LastBlock.Idx + 1,
--- a/gossip/c_block_callbacks.go
+++ b/gossip/c_block_callbacks.go
@@ -247,14 +247,14 @@ func consensusCallbackBeginBlockFn(
 
 				// skip LLR block/epoch deciding if not activated
 				if !es.Rules.Upgrades.Llr {
-					llrs := store.GetLlrState()
-					if llrs.LowestBlockToDecide == blockCtx.Idx {
-						llrs.LowestBlockToDecide++
-					}
-					if sealing && es.Epoch+1 == llrs.LowestEpochToDecide {
-						llrs.LowestEpochToDecide++
-					}
-					store.SetLlrState(llrs)
+					store.ModifyLlrState(func(llrs *LlrState) {
+						if llrs.LowestBlockToDecide == blockCtx.Idx {
+							llrs.LowestBlockToDecide++
+						}
+						if sealing && es.Epoch+1 == llrs.LowestEpochToDecide {
+							llrs.LowestEpochToDecide++
+						}
+					})
 				}
 
 				evmProcessor := blockProc.EVMModule.Start(blockCtx, statedb, evmStateReader, onNewLogAll, es.Rules)
--- a/gossip/store_llr_state.go
+++ b/gossip/store_llr_state.go
@@ -13,8 +13,16 @@ type LlrState struct {
 	LowestBlockToFill   idx.Block
 }
 
-func (s *Store) SetLlrState(llrs LlrState) {
+func (s *Store) setLlrState(llrs LlrState) {
 	s.cache.LlrState.Store(&llrs)
+}
+
+func (s *Store) ModifyLlrState(f func(*LlrState)) {
+	s.mutex.WriteLlrState.Lock()
+	defer s.mutex.WriteLlrState.Unlock()
+	llrs := s.GetLlrState()
+	f(&llrs)
+	s.setLlrState(llrs)
 }
 
 func (s *Store) GetLlrState() LlrState {
--- a/gossip/sync.go
+++ b/gossip/sync.go
@@ -150,9 +150,14 @@ func (h *handler) txsyncLoop() {
 }
 
 func (h *handler) updateSnapsyncStage() {
-	if h.config.AllowSnapsync && time.Since(h.store.GetEpochState().EpochStart.Time()) > snapsyncMinEndAge {
+	fullsyncPossible := h.store.evm.HasStateDB(h.store.GetBlockState().FinalizedStateRoot)
+	// never allow to stop fullsync as it may lead to a race condition due to overwritten EVM snapshot by snapsync
+	snapsyncPossible := h.config.AllowSnapsync && !h.syncStatus.Is(ssEvents)
+	snapsyncNeeded := !fullsyncPossible || time.Since(h.store.GetEpochState().EpochStart.Time()) > snapsyncMinEndAge
+
+	if snapsyncPossible && snapsyncNeeded {
 		h.syncStatus.Set(ssSnaps)
-	} else {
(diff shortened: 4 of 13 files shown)


gossip/apply_genesis.go
```diff
@@ -184,7 +184,7 @@ func (s *Store) applyEpoch1Genesis(blockProc BlockProc, g opera.Genesis) (err er
 	s.SetHistoryBlockEpochState(es.Epoch, bs, es)
 	s.SetEpochBlock(blockCtx.Idx+1, es.Epoch)
 	s.SetBlockEpochState(bs, es)
```
cyberbono3, line 186: Why we export `SetBlockEpochState`?  i just checked, it is used within `gossip` package only

gossip/c_block_callbacks.go
```diff
@@ -247,14 +247,14 @@ func consensusCallbackBeginBlockFn(
...
-					if llrs.LowestBlockToDecide == blockCtx.Idx {
-						llrs.LowestBlockToDecide++
-					}
-					if sealing && es.Epoch+1 == llrs.LowestEpochToDecide {
-						llrs.LowestEpochToDecide++
-					}
-					store.SetLlrState(llrs)
+					store.ModifyLlrState(func(llrs *LlrState) {
+						if llrs.LowestBlockToDecide == blockCtx.Idx {
+							llrs.LowestBlockToDecide++
```
cyberbono3, line 252: Can you explain why you increment `llrs.LowestBlockToDecide` here, please?

gossip/c_block_callbacks.go
```diff
@@ -247,14 +247,14 @@ func consensusCallbackBeginBlockFn(
...
+						}
+						if sealing && es.Epoch+1 == llrs.LowestEpochToDecide {
+							llrs.LowestEpochToDecide++
```
cyberbono3, line 255: and here?

gossip/store_llr_state.go
```diff
@@ -13,10 +13,18 @@ type LlrState struct {
...
 
-func (s *Store) SetLlrState(llrs LlrState) {
+func (s *Store) setLlrState(llrs LlrState) {
 	s.cache.LlrState.Store(&llrs)
 }
 
+func (s *Store) ModifyLlrState(f func(*LlrState)) {
+	s.mutex.WriteLlrState.Lock()
+	defer s.mutex.WriteLlrState.Unlock()
+	llrs := s.GetLlrState()
```
cyberbono3, line 23: `GetLlrState` should be private as well

gossip/sync.go
```diff
@@ -180,23 +185,32 @@ func (h *handler) snapsyncStageTick() {
...
 				continue
 			}
+			// cancel snapsync activity to prevent race condition
+			done := make(chan struct{})
+			h.snapState.updatesCh <- snapsyncStateUpd{
+				snapsyncCancelCmd: &snapsyncCancelCmd{done},
+			}
+			<-done
+			// finalize snapsync
 			err := h.process.SwitchEpochTo(epoch)
```
hadv, line 195: I prefer this code 
```suggestion
			if err := h.process.SwitchEpochTo(epoch); err != nil {
                h.Log.Error("Failed to result snapsync", "epoch", epoch, "block", bs.LastBlock.Idx, "err", err)			
			}
			h.Log.Info("snapsync is finalized at", "epoch", epoch, "block", bs.LastBlock.Idx, "root", bs.FinalizedStateRoot)
```

this info log message `Epoch is switched by snapsync` sound a bit mis-leading like snapsync is switched to sync at new state root. Can we update to `snapsync is finalized at` or smth like that?

gossip/c_block_callbacks.go
```diff
@@ -247,14 +247,14 @@ func consensusCallbackBeginBlockFn(
...
 				if !es.Rules.Upgrades.Llr {
-					llrs := store.GetLlrState()
-					if llrs.LowestBlockToDecide == blockCtx.Idx {
-						llrs.LowestBlockToDecide++
-					}
-					if sealing && es.Epoch+1 == llrs.LowestEpochToDecide {
-						llrs.LowestEpochToDecide++
-					}
-					store.SetLlrState(llrs)
+					store.ModifyLlrState(func(llrs *LlrState) {
```
hadv, line 250: this block `if !es.Rules.Upgrades.Llr {` means node always apply the **Llr** upgrade then it will rollback the Llr state if the node config doesn't support Llr upgrade, is it correct?

gossip/apply_genesis.go
uprendis: It's a part of the public interface - that's the only method to mutate `BlockEpochState`. We don't see most of gossip.Store method used elsewhere because only `gossip` processes consensus messages

gossip/c_block_callbacks.go
uprendis: Almost - indeed it will always enable the Llr upgrade if so was commanded by `driverpos.Topics.UpdateNetworkRules` (given that the node was upgraded when it processed the tx with the upgrade and knew what Llr upgrade is, otherwise it will have to upgrade and resync from scratch).
This `if` is needed so that node would correctly maintain `LowestBlockToDecide` and `LowestEpochToDecide` for period(s) when Llr is disabled.

After each processed block, this piece of code will increment LowestBlockToDecide, effectively signaling that the just processed block doesn't need any more LLR votes (since block doesn't belong to LLR period, it doesn't need LLR votes at all). The same for epochs.

For example:
Let's assume that LLR gets enabled at block 70. After LLR is enabled, LLR will not look for votes for blocks 1-69
Without this code, LLR would start to work with LowestBlockToDecide=1 and would infinitely search votes for blocks 1-69 (which don't exist because LLR validators don't vote for blocks before LLR)

gossip/c_block_callbacks.go
uprendis: Please see the reply above

gossip/c_block_callbacks.go
uprendis: Please see the reply above

gossip/sync.go
```diff
@@ -180,23 +185,32 @@ func (h *handler) snapsyncStageTick() {
...
 			if err == nil {
 				h.Log.Info("Epoch is switched by snapsync", "epoch", epoch, "block", bs.LastBlock.Idx, "root", bs.FinalizedStateRoot)
 			} else {
 				h.Log.Error("Failed to result snapsync", "epoch", epoch, "block", bs.LastBlock.Idx, "err", err)
 			}
+			h.syncStatus.Set(ssEvents)
```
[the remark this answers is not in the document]
uprendis: It's just a heuristic, there could be a different number of epochs in the loop (or just 1)
Imagine that we've just processed a new ER and the LowestEpochToFill is incremented and now 1001. Shortly after that, heal stage was finished for epoch 1000. Presence of this loop makes it possible to finalize snapsync at epoch 1000 even if it's not a latest epoch. 3 iterations means that we consider {latest LLR epoch, latest LLR epoch -1, latest LLR epoch -2} as candidates to finalize snapsync

hadv (approved):

gossip/apply_genesis.go
cyberbono3: Yes I get it what this method does. Since it is used only within `gossip` package, I would make  `SetBlockEpochState` private.

cyberbono3 (approved):
lgtm

gossip/apply_genesis.go
uprendis: If a method is used only in one package, it isn't yet sufficient to make it private. Private methods indicate that they are not a part of the "clean" interface and should be called with caution, `SetBlockEpochState` seems to be a part of the "clean" interface

uprendis: @rus-alex what do you think about it?

rus-alex: `gossip.Store` is enough independent and could be used as storage anywhere, not only by gossip package. `SetBlockEpochState()` should be public as part of clean interface and as pair method with public `GetBlockEpochState()`.

PR #222, opened 2022-01-24
No parallel snapshot generation

uprendis:
- allow events processing only after EVM snapshot generation has finished to prevent a non-deterministic result of transactions execution

### DIFF
--- a/gossip/sync.go
+++ b/gossip/sync.go
@@ -153,16 +153,17 @@ func (h *handler) txsyncLoop() {
 func (h *handler) updateSnapsyncStage() {
 	// never allow fullsync while EVM snap is still generating, as it may lead to a race condition
 	snapGenOngoing, _ := h.store.evm.Snaps.Generating()
-	fullsyncPossible := h.store.evm.HasStateDB(h.store.GetBlockState().FinalizedStateRoot) && !snapGenOngoing
+	fullsyncPossibleEver := h.store.evm.HasStateDB(h.store.GetBlockState().FinalizedStateRoot)
+	fullsyncPossibleNow := fullsyncPossibleEver && !snapGenOngoing
 	// never allow to stop fullsync as it may lead to a race condition due to overwritten EVM snapshot by snapsync
-	snapsyncPossible := h.config.AllowSnapsync && !h.syncStatus.Is(ssEvents)
-	snapsyncNeeded := !fullsyncPossible || time.Since(h.store.GetEpochState().EpochStart.Time()) > snapsyncMinEndAge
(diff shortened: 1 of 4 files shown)


hadv (approved):

cyberbono3 (approved):

gossip/sync.go
```diff
@@ -150,14 +151,19 @@ func (h *handler) txsyncLoop() {
 }
 
 func (h *handler) updateSnapsyncStage() {
-	fullsyncPossible := h.store.evm.HasStateDB(h.store.GetBlockState().FinalizedStateRoot)
+	// never allow fullsync while EVM snap is still generating, as it may lead to a race condition
+	snapGenOngoing, _ := h.store.evm.Snaps.Generating()
+	fullsyncPossibleEver := h.store.evm.HasStateDB(h.store.GetBlockState().FinalizedStateRoot)
+	fullsyncPossibleNow := fullsyncPossibleEver && !snapGenOngoing
```
hadv, line 157:
```suggestion
```

gossip/sync.go
```diff
@@ -150,14 +151,19 @@ func (h *handler) txsyncLoop() {
...
-	snapsyncNeeded := !fullsyncPossible || time.Since(h.store.GetEpochState().EpochStart.Time()) > snapsyncMinEndAge
+	snapsyncPossible := h.config.AllowSnapsync && (h.syncStatus.Is(ssUnknown) || h.syncStatus.Is(ssSnaps))
+	snapsyncNeeded := !fullsyncPossibleEver || time.Since(h.store.GetEpochState().EpochStart.Time()) > snapsyncMinEndAge
 
 	if snapsyncPossible && snapsyncNeeded {
 		h.syncStatus.Set(ssSnaps)
-	} else if fullsyncPossible {
+	} else if snapGenOngoing {
+		h.syncStatus.Set(ssEvmSnapGen)
+	} else if fullsyncPossibleNow {
```
hadv, line 166: I think that `fullsyncPossibleNow` is a bit redundant due to the `!snapGenOngoing` already checked in `else if snapGenOngoing` then `fullsyncPossibleEver` is enough. 

```suggestion
	} else if fullsyncPossibleEver {
```

hadv (changes requested):

gossip/sync.go
uprendis: Yeah, just wanted to make it clear that it's dangerous to launch `ssEvents` before snapshot generation is over, in a case if somebody will change the branches structure in future

hadv (approved):

PR #223, opened 2022-01-24
Fix mutation of cached HistoryBlockEpochState values

uprendis:
- fix race condition caused by mutation of cached BlockEpochState/HistoryBlockEpochState values
`.Copy()` method didn't copy the `Originated` field which gets mutated during block processing. As a result, cached value of HistoryBlockEpochState.BlockState gets mutated, which leads to retrieval of a different ER

### DIFF
--- a/gossip/store_decided_state.go
+++ b/gossip/store_decided_state.go
@@ -17,6 +17,7 @@ type BlockEpochState struct {
 }
 
 func (s *Store) SetHistoryBlockEpochState(epoch idx.Epoch, bs iblockproc.BlockState, es iblockproc.EpochState) {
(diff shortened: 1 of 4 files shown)


hadv:
Do you think we need to fix this Copy() as well?

https://github.com/Fantom-foundation/go-opera/blob/fa7e8da8d0f19e9294226ed8d9d8f102fde9f2b1/gossip/blockproc/sealmodule/sealer.go#L55

rus-alex (approved):

hadv (approved):

cyberbono3 (approved):
lgtm

PR #227, opened 2022-01-27
Fix inconsistent flush

uprendis:
- Fix bug introduced the #222 which allows DBs flushing without engineMu being locked. As a result, it may flush a dirty inconsistent state


hadv (approved):
Yes, we can see the comment on the first line of the functions 

```
// engineMu should be locked here
```

hadv (approved):

PR #230, opened 2022-01-31
Logs filter deduplication

rus-alex:
Prevents double result of "eth_getLogs" when client request parameter includes the same address or pattern twice.

### DIFF
--- a/topicsdb/topicsdb.go
+++ b/topicsdb/topicsdb.go
@@ -109,8 +109,11 @@ func limitPattern(pattern [][]common.Hash) (limited [][]common.Hash, err error) 
 	copy(limited, pattern)
 
 	ok := false
-	for _, variants := range limited {
+	for i, variants := range limited {
 		ok = ok || len(variants) > 0
+		if len(variants) > 1 {
+			limited[i] = uniqOnly(variants)
+		}
 	}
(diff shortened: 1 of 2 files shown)


topicsdb/topicsdb.go
```diff
@@ -120,6 +123,19 @@ func limitPattern(pattern [][]common.Hash) (limited [][]common.Hash, err error)
...
 }
 
+func uniqOnly(hh []common.Hash) []common.Hash {
+	index := make(map[common.Hash]struct{}, len(hh))
+	for _, h := range hh {
+		index[h] = struct{}{}
+	}
+
+	uniq := make([]common.Hash, 0, len(index))
+	for h := range index {
```
uprendis, line 133: Note: The `uniq` will be randomly sorted because map iteration is random. It seems to be okay, but please double check

rus-alex: Yes, that is why I use `require.ElementsMatch()` for test https://github.com/Fantom-foundation/go-opera/pull/230/files#diff-df517de37b0f47244945202c4b6243987c0b03adeb3392da5304d42a9d8c1fbdR393
I thinks it is OK, API client doesn't expects ordered logs list.

PR #233, opened 2022-02-01
Avoid parallel BRs and blocks processing

uprendis:
Don't allow BRs processing if events processing is allowed. For 2 reasons:
- avoid returning unprocessed blocks (but already indexed by BR processing) in API whose state is not accessible
- reduce number of DB writes, as block processing don't check if block is already written by BR processing and may overwrite the data. Since BR processing indexes every tx directly, while blocks processing indexes them via events, some txs may be indexed in both ways, which is a waste of disk storage

### DIFF
--- a/gossip/sync.go
+++ b/gossip/sync.go
@@ -55,6 +55,10 @@ func (ss *syncStatus) AcceptEvents() bool {
 	return ss.Is(ssEvents)
 }
 
+func (ss *syncStatus) AcceptBlockRecords() bool {
+	return !ss.Is(ssEvents)
+}
+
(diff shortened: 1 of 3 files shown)


cyberbono3 (approved):

gossip/sync.go
```diff
@@ -226,8 +230,10 @@ func (h *handler) snapsyncStageTick() {
 	// resume events downloading if events sync is enabled
 	if h.syncStatus.Is(ssEvents) {
```
hadv, line 231:
```suggestion
	if h.syncStatus.AcceptEvents() {
```

hadv (approved):
SGTM

PR #235, opened 2022-02-02
Fix kvdb snapshot switching race condition

uprendis:
- fix race condition in switchable.Snapshot
- add thread safety test for switchable.Snapshot

### DIFF
--- a/utils/switchable/snapshot_test.go
+++ b/utils/switchable/snapshot_test.go
@@ -1,0 +1,140 @@
+package switchable
+
+import (
+	"math/rand"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/Fantom-foundation/lachesis-base/common/bigendian"
+	"github.com/Fantom-foundation/lachesis-base/kvdb"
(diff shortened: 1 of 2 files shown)


hadv (approved):
LGTM

utils/switchable/snapshot_test.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func decodePair(b []byte) (uint32, uint32) {
+	v1 := bigendian.BytesToUint32(b[:4])
+	v2 := bigendian.BytesToUint32(b[4:])
+	return v1, v2
+}
+
+type UncallabaleAfterRelease struct {
```
rus-alex, line 22: How about https://github.com/Fantom-foundation/go-opera/commit/b689c69292de6a45b32d9eba3ed0b300b99e2731 addition ?

rus-alex (approved):

utils/switchable/snapshot_test.go
uprendis: Yeah sure, but no need in RW mutex, as it mutates the slice in both bases

PR #241, opened 2022-02-08
Save snapsync journal during stopping

uprendis:
- Save snapsync journal during stopping. Since it's saved only after `snapState.cancel()`, previously it was saved only during a switch of an epoch

### DIFF
--- a/gossip/sync.go
+++ b/gossip/sync.go
@@ -283,6 +283,10 @@ func (h *handler) snapsyncStateLoop() {
 				cmd.snapsyncCancelCmd.done <- struct{}{}
 			}
 		case <-h.snapState.quit:
+			if h.snapState.cancel != nil {
+				_ = h.snapState.cancel()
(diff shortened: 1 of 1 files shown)


gossip/sync.go
```diff
@@ -283,6 +283,10 @@ func (h *handler) snapsyncStateLoop() {
 				cmd.snapsyncCancelCmd.done <- struct{}{}
 			}
 		case <-h.snapState.quit:
+			if h.snapState.cancel != nil {
```
hadv, line 286: nit: I found that this block code appear 3 times in the `sync.go` file, so we should wrap it as a simple method like `cancelSnapsyncState()`?

```
				// cancel existing snapsync state
				if h.snapState.cancel != nil {
					_ = h.snapState.cancel()
					h.snapState.cancel = nil
				}
```

hadv (approved):

hadv (approved):

PR #247, opened 2022-02-09
[Config] Don't overwrite default dataDir from config file

peak3d:
Currently [Node] -> DataDir property specified in config TOML file gets overwritten when parsing command line parameters, even the command line parameter was not specified.

In case the commandline param is not given, the config already contains either the DataDir from TOML file or the default path.
There is no need to overwrite it in case the datadir commandline parameter is not given.

Steps for reproduction:
- Create a config.toml file containing:
```
[Node]
DataDir = "/tmp/test"
```
- launch opera --config config.toml dumpconfig


uprendis (approved):

hadv (approved):
Thanks, SGTM

cyberbono3 (approved):

rus-alex (approved):

PR #250, opened 2022-02-11
Fix nil pointer dereference in GetTransactionReceipt

uprendis:


ethapi/api.go
```diff
@@ -1623,13 +1623,9 @@ func (s *PublicTransactionPoolAPI) GetTransactionReceipt(ctx context.Context, ha
 		"type":              hexutil.Uint(tx.Type()),
 	}
 	// Assign the effective gas price paid
-	if !s.b.ChainConfig().IsLondon(bigblock) {
```
hadv, line 1626: that mean we cannot use the fork check in opera due to we enable it on net `rules` ?

uprendis:
I'll include this hotfix into 1.1.0-rc.3 to prevent the nil pointer dereference in API. Later on, we may need to refactor the chain config construction, so that it would store actual heights of EVM upgrades activation. We can implement it by maintaining a new table which will store heights of upgrades activation

PR #254, opened 2022-02-21
Fix panics on non-existing block RPC requests

hkalina:
Fix conditions in ethapi to fix logged panics like this:

```
ERROR[02-21|13:22:13.963] RPC method eth_getTransactionByHash crashed: runtime error: invalid memory address or nil pointer dereference
goroutine 211568 [running]:
github.com/ethereum/go-ethereum/rpc.(*callback).call.func1()
 	/home/opera/go/pkg/mod/github.com/!fantom-foundation/go-ethereum@v1.10.8-ftm-rc1.0.20211223105936-3db2e88af263/rpc/service.go:200 +0x89
 panic({0x13d2760, 0x21cbe90})
 	/usr/local/go/src/runtime/panic.go:1038 +0x215
 github.com/Fantom-foundation/go-opera/ethapi.(*PublicTransactionPoolAPI).GetTransactionByHash(0xc00e050540, {0x18465f8, 0xc0155001c0}, {0x6c, 0x71, 0x94, 0x68, 0xfc, 0xf5, 0xd1, ...})
	/home/opera/go/src/pebble-go-opera/ethapi/api.go:1553 +0xcd
```

### DIFF
--- a/ethapi/api.go
+++ b/ethapi/api.go
@@ -998,7 +998,7 @@ func DoEstimateGas(ctx context.Context, b Backend, args TransactionArgs, blockNr
 	// Recap the highest gas limit with account's available balance.
 	if feeCap.BitLen() != 0 {
 		state, _, err := b.StateAndHeaderByNumberOrHash(ctx, blockNrOrHash)
-		if err != nil {
+		if state == nil || err != nil {
 			return 0, err
 		}
 		balance := state.GetBalance(*args.From) // from can't be nil
@@ -1547,7 +1547,7 @@ func (s *PublicTransactionPoolAPI) GetTransactionByHash(ctx context.Context, has
 	}
 	if tx != nil {
 		header, err := s.b.HeaderByNumber(ctx, rpc.BlockNumber(blockNumber))
-		if err != nil {
+		if header == nil || err != nil {
 			return nil, err
 		}
 		return newRPCTransaction(tx, header.Hash, blockNumber, index, header.BaseFee), nil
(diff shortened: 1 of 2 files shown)


ethapi/api.go
```diff
@@ -998,7 +998,7 @@ func DoEstimateGas(ctx context.Context, b Backend, args TransactionArgs, blockNr
 	// Recap the highest gas limit with account's available balance.
 	if feeCap.BitLen() != 0 {
 		state, _, err := b.StateAndHeaderByNumberOrHash(ctx, blockNrOrHash)
-		if err != nil {
+		if state == nil || err != nil {
```
uprendis, line 1001: StateAndHeaderByNumberOrHash returns an error if not found, but it's okay to double check

uprendis (approved):

cyberbono3 (approved):

ethapi/api.go
```diff
@@ -1547,7 +1547,7 @@ func (s *PublicTransactionPoolAPI) GetTransactionByHash(ctx context.Context, has
 	}
 	if tx != nil {
 		header, err := s.b.HeaderByNumber(ctx, rpc.BlockNumber(blockNumber))
-		if err != nil {
+		if header == nil || err != nil {
```
hadv, line 1550: Basically it's okay to check this. 

Or it'd better to do it in other way by letting `HeaderByNumber()` will return `errors.New("xxx not found")` then we don't need to check **_nil_** anymore?

Similar to `BlockByNumber()` and other API methods.

hkalina: Current behavior (returning `nil, nil` from `HeaderByNumber`) is consistent with go-ethereum, for example:

https://github.com/ethereum/go-ethereum/blob/master/eth/api_backend.go#L75

If we change it, it can bring compatibility issues in porting code from go-ethereum.

hadv: okay, I got it.

Actually, go-ethereum is not consistent as well, `HeaderByNumberOrHash` may return `return nil, errors.New("header for hash not found")` as in 

https://github.com/ethereum/go-ethereum/blob/862f8e98bce89889de22e2d32bf37f823127a32c/eth/api_backend.go#L85

hadv (approved):

hkalina:
@uprendis @cyberbono3 @hadv Hello, I checked Contribution Guidelines - since this PR has review from more than two developers and I have no permission to merge it, can I ask you, reviewers, to merge this?

PR #264, opened 2022-03-07
Add new "fixdirty" tool for fixing database in dirty state

hkalina:
The PR introduce new tool, which resolves dirty state of database, which often occurs when Opera process is killed. (For example when the server is out of ram and the process is killed by OOM killer.)

This tool:
* removes epoch databases (which are created empty on a new epoch start) and lachesis database
* reverts the gossip database into state, in which it was at the end of the last closed epoch
* re-creates the lachesis database from the gossip database content
* cleans database dirty flags

This removes the necessity of removing the whole database and starting the sync from genesis again, which takes a week today. (The oncoming snap sync should help, but it cannot be used for archive nodes which should store the whole state history.)

### DIFF
--- a/cmd/opera/launcher/fixdirty.go
+++ b/cmd/opera/launcher/fixdirty.go
@@ -1,0 +1,199 @@
+package launcher
+
+import (
+	"fmt"
+	"path"
+	"strings"
+	"time"
+
+	"github.com/Fantom-foundation/lachesis-base/abft"
+	"github.com/Fantom-foundation/lachesis-base/common/bigendian"
+	"github.com/Fantom-foundation/lachesis-base/hash"
+	"github.com/Fantom-foundation/lachesis-base/inter/idx"
+	"github.com/Fantom-foundation/lachesis-base/kvdb"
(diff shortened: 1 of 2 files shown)


hkalina:
To be worked in:
> I would only add looping over recent epochs [lastEpoch-10000, lastEpoch] instead of trying only the previous epoch [lastEpoch-1], because there will be an addition of automatic pruning in future, and many epochs will have pruned MPTs

uprendis (approved):

rus-alex (approved):

hkalina:
@rus-alex Updated: refactored to fix the gossip database in standalone function (to close gdb with defer), sorted imports.

rus-alex (approved):

hkalina:
Successfully used to restore several RPC servers, which ended in dirty state after sigkill (probably from OOM killer). Servers are now in a valid state and respond to RPC requests correctly.

hkalina:
@uprendis @rus-alex  can this be merged?

PR #270, opened 2022-03-13
Fix service initialization during events import

uprendis:


hadv (approved):

cmd/opera/launcher/import.go
```diff
@@ -146,6 +146,11 @@ func importEventsFile(srv *gossip.Service, fn string) error {
 	signal.Notify(interrupt, syscall.SIGINT, syscall.SIGTERM)
 	defer signal.Stop(interrupt)
 
+	// wait until snapshot generation is complete
+	for srv.EvmSnapshotGeneration() {
+		time.Sleep(100 * time.Millisecond)
```
rus-alex, line 151: How about interruption here while snapshot generation?
```
select {
		case <-interrupt:
			return fmt.Errorf("interrupted")
		case <-time.After(100 * time.Millisecond):
		    continue
}
```

rus-alex (approved):

cmd/opera/launcher/import.go
uprendis: Unfortunately geth's `snapshot.Tree` interface doesn't have such a channel or a separate method for waiting until generation is finished

hadv: I think @rus-alex  mean we change the code like below then user still can `Ctrl-C ` to interrupt the import process while the snapshot is generating 

```
	// wait until snapshot generation is complete or be interrupted
	for srv.EvmSnapshotGeneration() {
		select {
		case <-interrupt:
			return fmt.Errorf("interrupted")
		case <-time.After(100 * time.Millisecond):
			continue
		}
	}
```

cyberbono3: > How about interruption here while snapshot generation?
> 
> ```
> select {
> 		case <-interrupt:
> 			return fmt.Errorf("interrupted")
> 		case <-time.After(100 * time.Millisecond):
> 		    continue
> }
> ```

What is the reason to implement interruption logic here?

cyberbono3: I don't know if it makes sense. I would modify `EvmSnapshotGeneration` to return bool channel. Then I would run `EvmSnapshotGeneration` in separate go routine and the main process should be blocked until `EvmSnapshotGeneration` bool channel returns true.

uprendis: > I think @rus-alex mean we change the code like below then user still can Ctrl-C to interrupt the import process while the snapshot is generating

Oh sorry, I get it now

rus-alex: Sorry for github UI unexpectedly allowed me to add commit to uprendis branch  o_0
Yes, I didn't mean to change the geth's interfaces.

hadv: > Sorry for github UI unexpectedly allowed me to add commit to uprendis branch o_0 Yes, I didn't mean to change the geth's interfaces.

Yes, we need to click and hold, drag mouse to the desired line number and release the mouse to make a code suggestion for multi lines of code

cyberbono3 (approved):
lgtm

PR #290, opened 2022-04-25
Fix of block proc metrics

rus-alex:
More accurate measurements of blocks processing metrics.

### DIFF
--- a/gossip/c_block_callbacks.go
+++ b/gossip/c_block_callbacks.go
@@ -257,7 +257,7 @@ func consensusCallbackBeginBlockFn(
 				}
 
 				evmProcessor := blockProc.EVMModule.Start(blockCtx, statedb, evmStateReader, onNewLogAll, es.Rules)
-				substart := time.Now()
+				startOfBlockExecution := time.Now()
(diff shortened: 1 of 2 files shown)


gossip/c_block_callbacks.go
```diff
@@ -311,11 +310,25 @@ func consensusCallbackBeginBlockFn(
 
 					_ = evmProcessor.Execute(txs, false)
 
+					startOfBlockWriting := time.Now()
+
 					evmBlock, skippedTxs, allReceipts := evmProcessor.Finalize()
```
hkalina, line 315: @rus-alex The `evmProcessor.Finalize` is included into the `blockWriterTimer`, but also into `blockExecutionTimer` - as it includes `statedb.Commit`, I suggest to include it only into `blockWriterTimer` - move the `blockExecutionTimer.Update` bellow before the `startOfBlockWriting`.

rus-alex: ok

rus-alex:
Outdated, see #358 and #363

PR #300, opened 2022-05-05
Add fee history API

uprendis:
Add a version of eth_feeHistory API call from web3 API. Due to the DAG consensus, estimating fee based on a history of recent blocks is a flawed approach. Thus, the method returns an estimation of `fee tip` depending on provided percentiles, instead of an actual history

### DIFF
--- a/ethapi/api.go
+++ b/ethapi/api.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Fantom-foundation/lachesis-base/inter/idx"
 	"github.com/davecgh/go-spew/spew"
 	"github.com/ethereum/go-ethereum/accounts"
 	"github.com/ethereum/go-ethereum/accounts/abi"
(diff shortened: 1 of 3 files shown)


ethapi/api.go
```diff
@@ -84,6 +87,103 @@ func (s *PublicEthereumAPI) MaxPriorityFeePerGas(ctx context.Context) (*hexutil.
...
+	}
+
+	tips := make([]*hexutil.Big, 0, len(rewardPercentiles))
+	for _, p := range rewardPercentiles {
+		ratio := tipScaleRatio(uint64(gasprice.DecimalUnit * p / 100.0))
+		scaledTip := scaleGasTip(goldTip, baseFee, ratio)
+		tips = append(tips, (*hexutil.Big)(scaledTip))
+	}
+	res.OldestBlock.ToInt().SetUint64(uint64(oldest))
+	for i := uint64(0); i < uint64(last-oldest+1); i++ {
```
hadv, line 179: we just keep it only to compatibility with EIP-1559 API but actually the values are the same for all blocks in range?

Issue #301, opened 2022-05-09
deploy official binary releases


AaronKutch:
It appears to me that only the source code artifacts are released. For the purposes of local and automated tests (In particular I work with a Github Actions CI that needs the binary 15 times in parallel), it would be great if there were a container I could `COPY` from on Dockerhub or a `.tar.gz` file of the binary with Github tags or on the official website. A clean build from source requires about a gigabyte of dependencies.

quan8:
Merged. Thanks.

PR #303, opened 2022-05-09
Release build with Actions CI

AaronKutch:
This is an example for what I want to solve #301 . Upon pushing a tag, this will include a `opera.tar.gz` artifact along with the release. There are other things we might want to do such as only releasing for `release/**` branches or including other options that `ncipollo/release-action@v1` has.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -44,8 +44,6 @@ require (
 	gopkg.in/urfave/cli.v1 v1.20.0
 )
 
+replace github.com/ethereum/go-ethereum => github.com/Fantom-foundation/go-ethereum v1.10.8-ftm-rc4
+
 replace github.com/dvyukov/go-fuzz => github.com/guzenok/go-fuzz v0.0.0-20210103140116-f9104dfb626f
-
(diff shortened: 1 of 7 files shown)


AaronKutch:
The Actions compiled binary appears to be needing `GLIBC_2.29` locally on my machine when it didn't when locally compiled but it does immediately work in the Docker container I was wanting

go.mod
```diff
@@ -44,5 +44,3 @@ require (
 )
 
 replace github.com/ethereum/go-ethereum => github.com/Fantom-foundation/go-ethereum v1.9.7-0.20210827160629-07563551b4c0
-
-replace github.com/dvyukov/go-fuzz => github.com/guzenok/go-fuzz v0.0.0-20210103140116-f9104dfb626f
```
rus-alex, line 48: Try the #309 instead of disable go module replacing.

AaronKutch:
This is an example of what it does https://github.com/AaronKutch/go-opera/releases/tag/onomy_release_0038cf8

rus-alex (approved):

PR #304, opened 2022-05-09
Rework exitwhensynced functionality

uprendis:
Add `--exitwhensynced.age` and `--exitwhensynced.epoch` flags instead of `--exitwhensynced` flag:
- `--exitwhensynced.age` may be used instead of `--exitwhensynced` to stop node once it syncs up to a relatively latest state. `--exitwhensynced.age` requires to specify an exact age threshold of what's considered a latest state (e.g. 5s)
- `--exitwhensynced.epoch` allows to stop node exactly when it reaches a certain epoch. Its intended usecase is genesis files generation for a certain epoch

### DIFF
--- a/cmd/opera/launcher/config.go
+++ b/cmd/opera/launcher/config.go
@@ -98,6 +98,14 @@ var (
 		Usage: `Blockchain sync mode ("full" or "snap")`,
 		Value: "full",
 	}
+	ExitWhenAgeFlag = cli.DurationFlag{
+		Name:  "exitwhensynced.age",
+		Usage: "Exits after synchronisation reaches the required age",
+	}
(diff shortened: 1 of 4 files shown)


hadv:
it sound a good feature, May we add more `--exitwhensynced.block` to stop node at a specific block height. it might be the use-cases for end-user due to account state is updated in block by block basis.

uprendis:
Unfortunately stopping node a specific block height wouldn't be trivial to implement. The atomic consensus message is event. Processing single event may lead to processing multiple blocks, so it's clear how to stop at some height >= condition but not exactly ==

cyberbono3 (approved):

hadv (approved):

PR #305, opened 2022-05-09
Rework version watcher

uprendis:
- version watches prevents node from processing new events if an upgrade was missed, and removes 2 options of either only logging about a missed upgrade activation (default) or halting the node on a missed upgrade activation (had to be explicitly enabled). This is to prevent situation when node needs to be resynced if an upgrade activation was missed and node halting on missed upgrade activation wasn't explicitly enabled

### DIFF
--- a/gossip/blockproc/verwatcher/version_watcher.go
+++ b/gossip/blockproc/verwatcher/version_watcher.go
@@ -11,12 +11,10 @@ import (
 	"github.com/Fantom-foundation/go-opera/logger"
 	"github.com/Fantom-foundation/go-opera/opera/contracts/driver"
 	"github.com/Fantom-foundation/go-opera/opera/contracts/driver/driverpos"
-	"github.com/Fantom-foundation/go-opera/utils/errlock"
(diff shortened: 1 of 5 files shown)


gossip/blockproc/verwatcher/version_watcher.go
```diff
@@ -11,71 +11,59 @@ import (
...
 		done:     make(chan struct{}),
 		Instance: logger.New(),
 	}
 }
 
+func (w *VerWarcher) Pause() error {
+	if w.store.GetNetworkVersion() > version.AsU64() {
+		return fmt.Errorf("Network upgrade %s was activated. Current node version is %s. "+
+			"Please upgrade your node to continue syncing.", version.U64ToString(w.store.GetNetworkVersion()), version.AsString())
+	} else if w.store.GetMissedVersion() > 0 {
```
hadv, line 37: Do we miss to call `SetMissedVersion()` somewhere?

cyberbono3 (approved):

hadv (approved):

PR #306, opened 2022-05-10
Extend Opera-specific API calls

uprendis:
- Add epoch number into header/block API output
- Add abft_getValidators API for retrieving validator weights and public keys
- Add ftm_getRules API call for retrieving network rules (e.g. throughput limits or upgrades activation status)
- Add ftm_getEpochBlock API call for retrieving block height in a beginning of an epoch

### DIFF
--- a/ethapi/api.go
+++ b/ethapi/api.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Fantom-foundation/lachesis-base/hash"
 	"github.com/Fantom-foundation/lachesis-base/inter/idx"
 	"github.com/davecgh/go-spew/spew"
 	"github.com/ethereum/go-ethereum/accounts"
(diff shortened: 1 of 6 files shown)


hadv (approved):
LGTM, it's a bit confuse when using `rpc.BlockNumber` for `epoch` but we already used it as it is before.

ethapi/api.go
```diff
@@ -680,6 +681,30 @@ func (s *PublicBlockChainAPI) CurrentEpoch(ctx context.Context) hexutil.Uint64 {
 	return hexutil.Uint64(s.b.CurrentEpoch(ctx))
 }
 
+// GetRules returns network rules for an epoch
+func (s *PublicBlockChainAPI) GetRules(ctx context.Context, epoch rpc.BlockNumber) (interface{}, error) {
```
cyberbono3, line 685: why `(s *PublicBlockChainAPI) GetRules` returns `(interface{}, error)` instead of `(opera.Rules, error)`?

cyberbono3 (approved):

PR #308, opened 2022-05-13
Separate genesis file into independent units

uprendis:
Genesis file:
- instead of using .zip container for storage of sections of genesis file, use custom gzip-based container, whose items can be separated or concatenated on binary level (i.e. units of a genesis file can split into independent files on binary level, and then re-arranged in an arbitrary order, as each unit is an independent piece with its own file header). It will help to reduce data redundancy on servers for hosting the genesis files, and will make it easier to incrementally generate diffs compared to prev genesis files rather than re-generating all data from scratch. Using gzip instead of zip is also advantageous for compression speed, because there's more performant parallel implementation for gzip parallel compression
- add dry mode for genesis generation, which calculates the data hashes without storing data or allocating disk space

Changes in filehash module:
- fix bug in filehash module which resulted into insecure hashing of genesis file (034bfedb22da3214b0b2a4fd9a4171d602be4757)
- add tests for filehash module


hadv (approved):

opera/genesisstore/fileshash/write_file.go
```diff
@@ -16,41 +18,57 @@ type TmpWriter interface {
...
 	}
 }
 
 func (w *Writer) writeIntoTmp(p []byte) error {
 	if len(p) == 0 {
 		return nil
 	}
-	if w.size/w.tmpSize >= uint64(len(w.tmps)) {
-		w.tmps = append(w.tmps, w.openTmp())
+	if w.size/w.pieceSize >= uint64(len(w.tmps)) {
```
cyberbono3, line 50: @uprendis can you explain the logic here, please?

uprendis: @cyberbono3 would you mean this change specifically, or `filehash` module in general?

PR #310, opened 2022-05-16
Check nil in topicsdb record fetch

thaarok:
This PR fixes panics occurring during eth_getLogs call:
```
May 14 00:27:53 xapi173 opera[1587273]: ERROR[05-14|00:27:53.892] RPC method eth_getLogs crashed: runtime error: slice bounds out of range [:32] with capacity 0
May 14 00:27:53 xapi173 opera[1587273]: goroutine 398114963 [running]:
May 14 00:27:53 xapi173 opera[1587273]: github.com/ethereum/go-ethereum/rpc.(*callback).call.func1()
May 14 00:27:53 xapi173 opera[1587273]:         /home/opera/go/pkg/mod/github.com/hkalina/go-ethereum@v1.9.7-0.20220420181000-8998b2a3f533/rpc/service.go:204 +0x89
May 14 00:27:53 xapi173 opera[1587273]: panic({0x15327a0, 0xc01be8a6d8})
May 14 00:27:53 xapi173 opera[1587273]:         /usr/local/go/src/runtime/panic.go:838 +0x207
May 14 00:27:53 xapi173 opera[1587273]: github.com/Fantom-foundation/go-opera/topicsdb.(*logrec).fetch(0xc0bdd88780, {0x7f14400c9570, 0xc002cf5630})
May 14 00:27:53 xapi173 opera[1587273]:         /home/opera/go/src/go-opera/topicsdb/record.go:47 +0x4f3
May 14 00:27:53 xapi173 opera[1587273]: github.com/Fantom-foundation/go-opera/topicsdb.(*Index).ForEachInBlocks.func1(0xc0bdd88780)
May 14 00:27:53 xapi173 opera[1587273]:         /home/opera/go/src/go-opera/topicsdb/topicsdb.go:91 +0x57
May 14 00:27:53 xapi173 opera[1587273]: github.com/Fantom-foundation/go-opera/topicsdb.(*Index).walkNexts(0xc08e5d7da0?, {0x188eca0?, 0xc09d52be60?}, 0x66?, {0xc00684c990?, 0x8?, 0x8?}, 0x5?, 0xc01bed8fe0?)
May 14 00:27:53 xapi173 opera[1587273]:         /home/opera/go/src/go-opera/topicsdb/search_lazy.go:94 +0x15a
[...]

PR #311, opened 2022-05-17
Fix of topicsdb RW-race

rus-alex:
Fix of read-write race #310 when log's index is already written and can be found but log's data is not yet.
Also includes:
 - test fix;
 - the meaning of max topics count;


hkalina:
I tried to deploy this on xapi173 where did the bug occurred and no panic occurred now - seems this fixes it correctly.

PR #314, opened 2022-05-19
Reactive gas price oracle

uprendis:
rework gasprice oracle to rely on two parts:
- Constructive part: increases gasprice according to accumulated gas power of validators (less -> higher price). This part is needed to protect network from draining validators. This part is needed because minGasPrice/baseFee balancing cannot be done very quickly per-block due to epoch-based consensus structure
- Reactive: takes gas price from top G's gas position in an average of txpool statistic for last 12 seconds, where G depends on chosen `certainty` (higher certainty -> smaller G). `Certainty` is a percentile of eth_feeHistory

Final gasprice estimation is a maximum of these 2 parts

Previously gasPrice estimation was based only on constructive part. This PR adjusts weights for the constructive part and improves calculation of accumulated gas power of validators

Note: like before, meeting the constructive part is not optional but mandatory, it serves as a soft minGasPrice/baseFee limit. This PR adds ftm_effectiveBaseFee API call for retrieving current soft limit for gas price

### DIFF
--- a/evmcore/tx_pool.go
+++ b/evmcore/tx_pool.go
@@ -139,7 +139,7 @@ type StateReader interface {
 	GetBlock(hash common.Hash, number uint64) *EvmBlock
 	StateAt(root common.Hash) (*state.StateDB, error)
 	MinGasPrice() *big.Int
-	RecommendedGasTip() *big.Int
+	EffectiveMinTip() *big.Int
 	MaxGasLimit() uint64
 	SubscribeNewBlock(ch chan<- ChainHeadNotify) notify.Subscription
 	Config() *params.ChainConfig
@@ -640,7 +640,7 @@ func (pool *TxPool) validateTx(tx *types.Transaction, local bool) error {
 		return ErrUnderpriced
 	}
 	// Ensure Opera-specific hard bounds
-	if recommendedGasTip, minPrice := pool.chain.RecommendedGasTip(), pool.chain.MinGasPrice(); recommendedGasTip != nil && minPrice != nil {
+	if recommendedGasTip, minPrice := pool.chain.EffectiveMinTip(), pool.chain.MinGasPrice(); recommendedGasTip != nil && minPrice != nil {
 		if tx.GasTipCapIntCmp(recommendedGasTip) < 0 {
 			return ErrUnderpriced
 		}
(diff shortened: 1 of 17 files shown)


hadv (approved):
LGTM

evmcore/tx_pool.go
```diff
@@ -555,6 +555,17 @@ func (pool *TxPool) Pending(enforceTips bool) (map[common.Address]types.Transact
 	return pending, nil
 }
 
+func (pool *TxPool) PendingSlice() types.Transactions {
```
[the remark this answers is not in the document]
uprendis: Oh it's just because there's already a method called `Pending` which returns a map, I wanted to emphasise the difference between them

rus-alex (changes requested):
Not all the PendingSlice() renamed to PendingTxs(). Test failed

evmcore/tx_pool.go
```diff
@@ -555,6 +555,17 @@ func (pool *TxPool) Pending(enforceTips bool) (map[common.Address]types.Transact
...
+	pool.mu.Lock()
+	defer pool.mu.Unlock()
+
+	pending := make(types.Transactions, 0, 1000)
```
cyberbono3, line 562: I think it would be better to set `len(pool.pending)` instead of `1000` in `pending`

cyberbono3, line 562: are you sure, that 1000 is the sufficient upper bound for pending items?

hadv: @cyberbono3 I think that it's only the initialize capacity, slice will grow if it need more when using `append` func

cyberbono3 (approved):

cyberbono3:
I suggest to increase test coverage of `gossip/gasprice` package
```ok  	github.com/Fantom-foundation/go-opera/gossip/gasprice	0.148s	coverage: 0.0% of statements [no tests to run]```

uprendis:
@cyberbono3 added tests in [e4ddb6d](https://github.com/Fantom-foundation/go-opera/pull/314/commits/e4ddb6d6606998c5f8f2c1a4d85ab34b9bdb7997)

BMItr:
👌  great work

PR #328, opened 2022-06-14
Checking build at each commit using GitHub Actions

kirillt:
Hello, this is really small PR, but I think it would be very useful for wider adoption.

E.g. I'm experimenting with different networks and capability to download latest or some fixed binary build would really ease my life since then I can just take any fresh cloud instance, download the binary using my scripts and run my tests. I haven't used your template since it isn't a package and nothing really changed in the code itself. Only publishing archive with fresh binary build.

Please, let me know if this can be merged and what changes it needs for this. We can limit building for particular versions although I think it is useful to check build every commit.


kirillt:
Replaced by https://github.com/Fantom-foundation/go-opera/pull/329

PR #329, opened 2022-06-14
Checking build at every commit and releasing tags

kirillt:
Hello, this is really small PR, but I think it would be very useful for wider adoption.

E.g. I'm experimenting with different networks and capability to download binary build would really ease my life since, in this case, I could just take any fresh cloud instance, download the binary using my scripts and run my tests. No need to install Go on target machine and compile stuff from scratch (which isn't hard but takes time especially if it's automated testing environment). I haven't used your template since it isn't a package and nothing really changed in the code itself. Only publishing archive with fresh binary build.

Please, let me know if this can be merged and what changes it needs for this. We can limit building for particular versions although I think it is useful to check build every commit.

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -1,12 +1,9 @@
-name: Release
+name: Check build
 
-on:
-  push:
-    tags:
-    - '*'
+on: [push]
 
 jobs:
-  release:
+  check-build:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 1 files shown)


kirillt:
Testing builds are provided as job artifacts and those are downloadable only by authenticated users (normally from browsers). Such artifacts are packed as ZIP archives and weigh ~ 10MB.

Release builds are downloadable by anybody (possible to `wget` it right on a server) but they are naked binaries and weigh ~ 28MB. Possible improvement could be packing them before releasing. Another minor enhancement would be preservation of permissions, so it would be not needed to do `chmod +x opera` before running it. But GitHub seems to make it difficult somehow.

It seems that MIT license doesn't require packing sources together with binary release, but please fix me if I'm wrong.

PR #333, opened 2022-06-20
Add zero block into API

uprendis:
- add fake zero genesis block for compatibility with web3 API

### DIFF
--- a/gossip/evm_state_reader.go
+++ b/gossip/evm_state_reader.go
@@ -2,7 +2,6 @@ package gossip
 
 import (
 	"math/big"
-	"time"
 
 	"github.com/Fantom-foundation/lachesis-base/hash"
 	"github.com/Fantom-foundation/lachesis-base/inter/idx"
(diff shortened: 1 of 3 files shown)


gossip/evm_state_reader.go
```diff
@@ -88,6 +90,13 @@ func (r *EvmStateReader) GetBlock(h common.Hash, n uint64) *evmcore.EvmBlock {
 }
 
 func (r *EvmStateReader) getBlock(h hash.Event, n idx.Block, readTxs bool) *evmcore.EvmBlock {
+	if n == 0 {
+		// fake genesis block for compatibility with web3
+		return evmcore.NewEvmBlock(evmcore.ToEvmHeader(&inter.Block{
+			Time:    inter.Timestamp(1608600000 * time.Second),
```
hadv, line 96: nitpick: we can reuse `FakeGenesisTime` from **evmcore**
```suggestion
			Time:    evmcore.FakeGenesisTime,
```

uprendis: It would create a dependency cycle unfortunately

hadv: I found that we already imported the **evmcore** package for `evmcore.NewEvmBlock`

uprendis: Oh yes, you're right

hadv (approved):

PR #337, opened 2022-07-05
fix of default bootnodes value

rus-alex:
Special (one empty node) value of bootnodes is using to determine when we should use network default values.
But `go-ethereum/p2p/enode.Node` can't parse this value because of empty is not valid:
```
./opera dumpconfig > example.cfg
./opera --config=example.cfg
Fatal: TOML config file error: 1.1.1-rc.1.cfg, line 17: (p2p.Config.BootstrapNodes) EOF.
Use 'dumpconfig' command to get an example config file.

```
PR detects empty node before parsing, on AST stage.

The others way to fix it were more complicated in terms of maintain (IMHO):

1. Replace empty node with fake node (it was implemented as fast solution);
2. Modify `ethereum/go-ethereum/p2p/enode` to allow empty node parsing;
3. Use modified copy of config struct inside config.UnmarshalTOML();

### DIFF
--- a/cmd/opera/launcher/config_custom.go
+++ b/cmd/opera/launcher/config_custom.go
@@ -1,0 +1,79 @@
+package launcher
+
+import (
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/p2p/enode"
+	"github.com/naoina/toml"
+	"github.com/naoina/toml/ast"
+)
+
+// asDefault is slice with one empty element
+// which indicates that network default bootnodes should be substituted
+var asDefault = []*enode.Node{{}}
+
+func needDefaultBootnodes(nn []*enode.Node) bool {
+	return len(nn) == len(asDefault) && nn[0] == asDefault[0]
+}
+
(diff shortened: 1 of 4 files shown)


cmd/opera/launcher/config_custom.go
```diff
@@ -0,0 +1,79 @@
...
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/p2p/enode"
+	"github.com/naoina/toml"
+	"github.com/naoina/toml/ast"
+)
+
+// asDefault is slice with one empty element
+// which indicates that network default bootnodes should be substituted
+var asDefault = []*enode.Node{{}}
```
hadv, line 13: nitpick: I think the `asDefault` also be used at https://github.com/Fantom-foundation/go-opera/blob/master/cmd/opera/launcher/config.go#L371 then I thought that we would keep it in `params.go` as it is as before

hadv (approved):
basically LGTM, just a nitpick comment

PR #338, opened 2022-07-07
Forbid unlocking accounts with enabled external RPC

uprendis:
- fix initialization of ExtRPCEnabled var so that unlocking of accounts is forbidden unless `--allow-insecure-unlock` argument is added

### DIFF
--- a/gossip/ethapi_backend.go
+++ b/gossip/ethapi_backend.go
@@ -39,6 +39,11 @@ type EthAPIBackend struct {
 	allowUnprotectedTxs bool
 }
 
+// SetExtRPCEnabled updates extRPCEnabled
+func (b *EthAPIBackend) SetExtRPCEnabled(v bool) {
+	b.extRPCEnabled = v
+}
+
 // ChainConfig returns the active chain configuration.
 func (b *EthAPIBackend) ChainConfig() *params.ChainConfig {
 	return b.svc.store.GetEvmChainConfig()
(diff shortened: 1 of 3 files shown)


gossip/service.go
```diff
@@ -269,7 +270,7 @@ func newService(config Config, store *Store, blockProc BlockProc, engine lachesi
 	}
 
 	// create API backend
-	svc.EthAPI = &EthAPIBackend{config.ExtRPCEnabled, svc, stateReader, txSigner, config.AllowUnprotectedTxs}
+	svc.EthAPI = &EthAPIBackend{false, svc, stateReader, txSigner, config.AllowUnprotectedTxs}
```
hadv, line 273: it seems that we intend to init this with `true`? otherwise, it still work like before

```suggestion
	svc.EthAPI = &EthAPIBackend{true, svc, stateReader, txSigner, config.AllowUnprotectedTxs}
```

gossip/ethapi_backend.go
```diff
@@ -39,6 +39,11 @@ type EthAPIBackend struct {
 	allowUnprotectedTxs bool
 }
 
+// SetExtRPCEnabled updates extRPCEnabled
+func (b *EthAPIBackend) SetExtRPCEnabled(v bool) {
```
rus-alex, line 43: Setter is not needed, just assign it directly.

rus-alex (approved):
Note, incompatible with previous dumped configs:
`TOML config file error: field corresponding to "ExtRPCEnabled" is not defined in gossip.Config`.

gossip/service.go
uprendis: It's initialized by SetExtRPCEnabled later

gossip/ethapi_backend.go
uprendis: I just wanted to highlight that this variable may be initialized after object creation

hadv (approved):

cyberbono3 (approved):
lgtm

PR #344, opened 2022-07-26
fix nil cachescale issue after loading toml config file

hadv:
Set **cachescale** to default `cacheRatio` after loading config file to avoid nil pointer exception when running node with toml config file  

It seems the side effect of https://github.com/Fantom-foundation/go-opera/pull/337


uprendis:
It seems this issue is fixed in #343 by erasing cachescale from this struct

hadv:
> It seems this issue is fixed in #343 by erasing cachescale from this struct

okay, I need to run test the code that base on current `develop` branch and facing the issue. If we merge the #343 then we don't need this PR anymore. I just use as local for testing only.

PR #347, opened 2022-08-02
update some metrics for multi-db

hadv:


hadv:
This PR depends on https://github.com/Fantom-foundation/lachesis-base/pull/39

hadv:
I will setup grafana dashboard again to check the metrics on the latest `develop` branch

hadv:
I need to investigate a little bit more, still don't know why with the **multi-db**, there're brunch of metrics for a single db name, for instance, for the `gossip` in the below screen shot

Screen Shot 2022-08-02 at 18 52 53

uprendis:
Each DB table is treated as a separated DB now which has to be opened and closed. Behind the hood, multidb producer routs virtual DB names like `gossip/#` or `evm/M` to physical DBs and prefixes according to its config. For example, by default, `gossip/#` gets routed to DB with type=pebble-fsh, name=main, prefix=#

uprendis:
If we need metrics for each physical DB, then it would be better to install the wrapper for physical DB producers here: https://github.com/Fantom-foundation/go-opera/blob/11c9b85cbfe536469489626cda452cd7c739deeb/integration/db.go#L121

hadv:
> If we need metrics for each physical DB, then it would be better to install the wrapper for physical DB producers here:
> 
> https://github.com/Fantom-foundation/go-opera/blob/11c9b85cbfe536469489626cda452cd7c739deeb/integration/db.go#L121

Thank @uprendis I will do it

hadv:
So with multi-db we will has some separate metrics for each db like below

Screen Shot 2022-09-01 at 16 43 54

The good thing is we can monitor the data read/write metric more details.

By the way, it'd be good if we can sum the metrics on **grafana** dashboard for all dbs into single metric then we can it in total. I tried to use the transformations of the **grafana** but it still not is what we want. I'm finding another way.

PR #348, opened 2022-08-02
Fix block processing metrics, remove unused metrics

hkalina:
Fix block processing metrics:
* remove validation mentions (no validations happens in here, copypasta from go-ethereum)
* blockWriteTimer measured since the commit start (there have been missing `substart` assignment - replaced by two variables `executionStart`/`commitStart` to prevent similar future issues)


hkalina:
@uprendis

hadv (approved):
LGTM

PR #350, opened 2022-08-07
Desyncronize flushes

uprendis:
- de-synchronize flushes between nodes by randomizing flushing threshold for every epoch


gossip/service.go
```diff
@@ -310,13 +310,13 @@ func (s *Service) makePeriodicFlusher() PeriodicFlusher {
 			commitNeeded: func() bool {
 				// use slightly higher size threshold to avoid locking the mutex/wg pair and hurting events/blocks concurrency
```
cyberbono3, line 311: what kind of locking mechanism you mean here? can you clarify please?

uprendis: Originally, there was single mutex `engineMu` for events processing, including blocks processing. Eventually, blocks processing was partially evicted into a separate goroutine to allow parallel processing of up to one event and up to one block. It involved addition of one more synchronization mechanism - the wait group `blockProcWg`. When some actor (periodic flusher in this case) locks `engineMu` and waits for `blockProcWg`, it does it not atomically but sequentially - there's a period when `engineMu` is already locked but `blockProcWg` is being waited for. During this period, events processor has to wait for `engineMu` and cannot process new events

uprendis: Higher threshold here is intended so that most flushes would be done at the end of events processing at
```
s.mayCommit(newEpoch != oldEpoch)
```
when `engineMu` is already locked anyway

cyberbono3 (approved):

PR #357, opened 2022-08-29
Allow to preserve legacy DBs layout, use leveldb-only layout by default

uprendis:
- add presets for DBs layout: ldb-1 (fastest leveldb-only layout), pbl-1 (fastest pebble/leveldb layout), legacy-ldb (legacy layout from 1.1.1 or earlier), legacy-pbl (layout from `1.1.1-rc.2-pebble`). ldb-1 is used by default instead of pbl-1.
- add `reformat` mode for multidb migration for instant migration with unaltered DBs layout

Migration options:
- Instant migration from a leveldb-based version: `--db.migration.mode reformat --db.preset legacy-ldb` 
- Instant migration from 1.1.1-rc.2-pebble: `--db.migration.mode reformat --db.preset legacy-pbl` 
- Migration with rebuilding DBs according to any configured layout: `--db.migration.mode rebuild`

Flag `--db.migration.mode` is required only if datadir isn't migrated.

### DIFF
--- a/cmd/opera/launcher/config.go
+++ b/cmd/opera/launcher/config.go
@@ -111,6 +111,15 @@ var (
 		Name:  "exitwhensynced.epoch",
 		Usage: "Exits after synchronisation reaches the required epoch",
 	}
+
+	DBMigrationModeFlag = cli.StringFlag{
+		Name:  "db.migration.mode",
+		Usage: "MultiDB migration mode ('reformat' or 'rebuild')",
+	}
+	DBPresetFlag = cli.StringFlag{
+		Name:  "db.preset",
+		Usage: "DBs layout preset ('pbl-1' or 'ldb-1' or 'legacy-ldb' or 'legacy-pbl')",
+	}
 )
 
 type GenesisTemplate struct {
@@ -336,6 +345,27 @@ func gossipStoreConfigWithFlags(ctx *cli.Context, src gossip.StoreConfig) (gossi
 	return cfg, nil
 }
 
+func setDBConfig(ctx *cli.Context, cfg integration.DBsConfig, cacheRatio cachescale.Func) integration.DBsConfig {
+	if ctx.GlobalIsSet(DBPresetFlag.Name) {
+		preset := ctx.GlobalString(DBPresetFlag.Name)
+		if preset == "pbl-1" {
+			cfg = integration.Pbl1DBsConfig(cacheRatio.U64, uint64(utils.MakeDatabaseHandles()))
+		} else if preset == "ldb-1" {
+			utils.Fatalf("--%s: ldb-1 preset is not unsupported yet", DBPresetFlag.Name)
+		} else if preset == "legacy-ldb" {
+			cfg = integration.LdbLegacyDBsConfig(cacheRatio.U64, uint64(utils.MakeDatabaseHandles()))
+		} else if preset == "legacy-pbl" {
+			cfg = integration.PblLegacyDBsConfig(cacheRatio.U64, uint64(utils.MakeDatabaseHandles()))
+		} else {
+			utils.Fatalf("--%s must be 'pbl-1', 'ldb-1', 'legacy-pbl' or 'legacy-ldb'", DBPresetFlag.Name)
+		}
+	}
+	if ctx.GlobalIsSet(DBMigrationModeFlag.Name) {
+		cfg.MigrationMode = ctx.GlobalString(DBMigrationModeFlag.Name)
+	}
+	return cfg
+}
+
 func nodeConfigWithFlags(ctx *cli.Context, cfg node.Config) node.Config {
 	utils.SetNodeConfig(ctx, &cfg)
 
@@ -411,6 +441,7 @@ func mayMakeAllConfigs(ctx *cli.Context) (*config, error) {
 		return nil, err
 	}
 	cfg.Node = nodeConfigWithFlags(ctx, cfg.Node)
+	cfg.DBs = setDBConfig(ctx, cfg.DBs, cacheRatio)
 
 	err = setValidator(ctx, &cfg.Emitter)
 	if err != nil {
--- a/cmd/opera/launcher/config.go
+++ b/cmd/opera/launcher/config.go
@@ -351,7 +351,7 @@ func setDBConfig(ctx *cli.Context, cfg integration.DBsConfig, cacheRatio cachesc
 		if preset == "pbl-1" {
 			cfg = integration.Pbl1DBsConfig(cacheRatio.U64, uint64(utils.MakeDatabaseHandles()))
 		} else if preset == "ldb-1" {
-			utils.Fatalf("--%s: ldb-1 preset is not unsupported yet", DBPresetFlag.Name)
+			cfg = integration.Ldb1DBsConfig(cacheRatio.U64, uint64(utils.MakeDatabaseHandles()))
 		} else if preset == "legacy-ldb" {
 			cfg = integration.LdbLegacyDBsConfig(cacheRatio.U64, uint64(utils.MakeDatabaseHandles()))
 		} else if preset == "legacy-pbl" {
--- a/cmd/opera/launcher/config.go
+++ b/cmd/opera/launcher/config.go
@@ -348,15 +348,16 @@ func gossipStoreConfigWithFlags(ctx *cli.Context, src gossip.StoreConfig) (gossi
 func setDBConfig(ctx *cli.Context, cfg integration.DBsConfig, cacheRatio cachescale.Func) integration.DBsConfig {
 	if ctx.GlobalIsSet(DBPresetFlag.Name) {
 		preset := ctx.GlobalString(DBPresetFlag.Name)
-		if preset == "pbl-1" {
+		switch preset {
+		case "pbl-1":
 			cfg = integration.Pbl1DBsConfig(cacheRatio.U64, uint64(utils.MakeDatabaseHandles()))
-		} else if preset == "ldb-1" {
+		case "ldb-1":
 			cfg = integration.Ldb1DBsConfig(cacheRatio.U64, uint64(utils.MakeDatabaseHandles()))
-		} else if preset == "legacy-ldb" {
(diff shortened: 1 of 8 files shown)


hkalina:
I think Opera should throw an error when trying to use `reformat` mode with non-legacy preset.

Currently it continues the reformat and fails on broken database:
```
opera --datadir=/var/opera/mainnet --db.preset=ldb-1 --db.migration.mode=reformat
INFO [08-29|20:31:46.808] Maximum peer count                       total=50
INFO [08-29|20:31:46.808] Smartcard socket not found, disabling    err="stat /run/pcscd/pcscd.comm: no such file or directory"
CRIT [08-29|20:31:47.013] LLR state reading failed: genesis not applied
```

integration/legacy_migrate.go
```diff
@@ -186,7 +186,7 @@ func translateGossipPrefix(p byte) byte {
 	return p
 }
 
-func migrateLegacyDBs(chaindataDir string, dbs kvdb.FlushableDBProducer) error {
+func migrateLegacyDBs(chaindataDir string, dbs kvdb.FlushableDBProducer, mode string) error {
 	if !isEmpty(path.Join(chaindataDir, "gossip")) {
```
hadv, line 190: I think this check is not much necessary due to we already check it before calling the function `migrateLegacyDBs()`; it's okay to re-check inside the function but I prefer to use short circuit as below to reduce one long nested if condition

```suggestion
	if isEmpty(path.Join(chaindataDir, "gossip")) {
	     return null
	}
```

hadv:
> I think Opera should throw an error when trying to use `reformat` mode with non-legacy preset.
> 
> Currently it continues the reformat and fails on broken database:
> 
> ```
> opera --datadir=/var/opera/mainnet --db.preset=ldb-1 --db.migration.mode=reformat
> INFO [08-29|20:31:46.808] Maximum peer count                       total=50
> INFO [08-29|20:31:46.808] Smartcard socket not found, disabling    err="stat /run/pcscd/pcscd.comm: no such file or directory"
> CRIT [08-29|20:31:47.013] LLR state reading failed: genesis not applied
> ```

I think it only migrate the legacy preset via this checking? if legacy preset is presented then the `gossip` folder does not existed; but of course user can force the `--db.preset=ldb-1` to a legacy preset then it might be the issue that you mentioned.

```
if !isEmpty(path.Join(chaindataDir, "gossip")) {
```

hadv:
> Instant migration from a leveldb-based version: --db.migration.mode reformat --db.preset legacy-ldb
> Instant migration from 1.1.1-rc.2-pebble: --db.migration.mode reformat --db.preset legacy-pbl

If I understand correctly then the from preset layout in the case of `reformat` is deciding by below code, not by the `--db.preset`?

```
		if fileExists(path.Join(chaindataDir, "gossip", "LOG")) {
			oldDBs = leveldb.NewProducer(chaindataDir, cacheFn)
			oldDBsType = "ldb"
		} else {
			oldDBs = pebble.NewProducer(chaindataDir, cacheFn)
			oldDBsType = "pbl"
		}
```

uprendis:
```
so the --db.preset will be only meaning used for target preset layout in --db.migration.mode rebuild, right? And what happen if the target preset is legacy one?
```
That's right, ``reformat` and configured DB presets are technically independent from each other. But practically, every DB has two hidden records:
- `FlushIDKey` -> FlushID (it was always present)
- `TablesKey` -> list of tables of this DB (new record)
`reformat` currently just moves the tables without adding `TablesKeys` records. On first launch, multidb will think that those DBs are empty and will add `TablesKeys` itself according to configured layout. If configured layout is wrong, then it'll place wrong `TablesKeys` records which will confuse multidb to a point when datadir cannot be recovered anymore

So I have to either:
- Not just rename DBs but place correct `TablesKeys` records in there. In a case you set up a wrong DB layout, then it will just say that configured layout is different than datadir's and will exit without "damaging" the datadir
- Just hardcode that if `reformat` flag is present, then configured DB layout should strictly match to either legacy-ldb or legacy-pbl, depending on datadir's DB types

uprendis:
```
Just hardcode that if reformat flag is present, then configured DB layout should strictly match to either legacy-ldb or legacy-pbl, depending on datadir's DB types
```
Implemented this in [dff7354](https://github.com/Fantom-foundation/go-opera/pull/357/commits/dff7354d884f5889adb1f03e9d5de6e5cc970055)

hadv (approved):
LGTM

cmd/opera/launcher/config.go
```diff
@@ -336,6 +345,27 @@ func gossipStoreConfigWithFlags(ctx *cli.Context, src gossip.StoreConfig) (gossi
...
+			cfg = integration.Pbl1DBsConfig(cacheRatio.U64, uint64(utils.MakeDatabaseHandles()))
+		} else if preset == "ldb-1" {
+			cfg = integration.Ldb1DBsConfig(cacheRatio.U64, uint64(utils.MakeDatabaseHandles()))
+		} else if preset == "legacy-ldb" {
+			cfg = integration.LdbLegacyDBsConfig(cacheRatio.U64, uint64(utils.MakeDatabaseHandles()))
+		} else if preset == "legacy-pbl" {
+			cfg = integration.PblLegacyDBsConfig(cacheRatio.U64, uint64(utils.MakeDatabaseHandles()))
+		} else {
+			utils.Fatalf("--%s must be 'pbl-1', 'ldb-1', 'legacy-pbl' or 'legacy-ldb'", DBPresetFlag.Name)
+		}
```
rus-alex, line 361:
```suggestion
		switch preset {
		case "pbl-1":
			cfg = integration.Pbl1DBsConfig(cacheRatio.U64, uint64(utils.MakeDatabaseHandles()))
		case "ldb-1":
			cfg = integration.Ldb1DBsConfig(cacheRatio.U64, uint64(utils.MakeDatabaseHandles()))
		case "legacy-ldb":
			cfg = integration.LdbLegacyDBsConfig(cacheRatio.U64, uint64(utils.MakeDatabaseHandles()))
		case "legacy-pbl":
			cfg = integration.PblLegacyDBsConfig(cacheRatio.U64, uint64(utils.MakeDatabaseHandles()))
		default:
			utils.Fatalf("--%s must be 'pbl-1', 'ldb-1', 'legacy-pbl' or 'legacy-ldb'", DBPresetFlag.Name)
		}
```

integration/legacy_migrate.go
```diff
@@ -221,67 +236,105 @@ func migrateLegacyDBs(chaindataDir string, dbs kvdb.FlushableDBProducer) error {
...
-					openSrc: func() kvdb.Store {
-						return skipkeys.Wrap(openOldDB(name), MetadataPrefix)
-					},
-					openDst: func() kvdb.Store {
-						return openNewDB(name)
-					},
-					name: name,
-					dir:  chaindataDir,
-				})
+		if mode == "rebuild" {
```
rus-alex, line 239:
```suggestion
		switch mode {
		case "rebuild":
```

integration/legacy_migrate.go
```diff
@@ -221,67 +236,105 @@ func migrateLegacyDBs(chaindataDir string, dbs kvdb.FlushableDBProducer) error {
...
+						} else {
+							return openNewDB("gossip/" + string([]byte{translateGossipPrefix(byte(b))}))
+						}
+					},
+					name:    fmt.Sprintf("gossip/%c", rune(b)),
+					dir:     chaindataDir,
+					dropSrc: b == 0xff,
+				})
+			}
+		} else if mode == "reformat" {
```
rus-alex, line 302:
```suggestion
		case "reformat":
```

integration/legacy_migrate.go
```diff
@@ -221,67 +236,105 @@ func migrateLegacyDBs(chaindataDir string, dbs kvdb.FlushableDBProducer) error {
...
+				for _, name := range oldDBs.Names() {
+					if strings.HasPrefix(name, "lachesis") || strings.HasPrefix(name, "gossip-") {
+						err = os.Rename(path.Join(chaindataDir, name), path.Join(chaindataDir, "pebble-fsh", name))
+						if err != nil {
+							return err
+						}
+					}
+				}
+			}
+		} else {
```
rus-alex, line 336:
```suggestion
		default:
```

rus-alex (approved):

keennay:
Hey team, I have a question here. Also, if there's a more appropriate medium to ask besides this comments section, please let me know.

I am attempting to run a fresh archive node sync using txtracing/1.1.2-rc.6 on PebbleDB, while also using genesis `mainnet-109331-full-mpt.g` to retain archival data. Once my node finishes initializing with the archive genesis file, would I still need to rebuild the database for this v1.1.2-rc6 release? I'm not exactly sure if initializing the genesis file with the PebbleDB flags automatically converts all historical states to a PebbleDB format, but I've been running a node for a while & wanted to ensure everything was done properly.

I'm running an archive node today with the following flags:

`/blockchain/clients/ftm/go-opera/build/opera --genesis=/blockchain/nodes/ftm/go-opera/mainnet/mainnet-109331-full-mpt.g --syncmode=full --gcmode=archive --db.preset=pbl-1 --port=30305 --http --http.vhosts="*" --http.corsdomain="*" --http.addr=0.0.0.0 --http.port=8545 --ws --ws.origins="*" --ws.addr=0.0.0.0 --ws.port=8546 --datadir=/blockchain/nodes/ftm/go-opera/mainnet --tracenode --tracing --http.api="ftm,eth,debug,admin,web3,personal,net,txpool,sfc"`

Thanks,
Yannick

jmpike:
TL;DR: If you used 1.1.2-rc.6 with Pebble to initialise the DB from genesis, you don't need to do anything else.

The genesis file contains data needed to start a node, in this case including full state history, but in a general form. It doesn’t really convey  the low level data storage format, just the keys and values needed to reconstruct the desired state. So if you use it to initialise a node, it will use your DB backend setting and layout preset even for the data state reconstruction at the beginning of the process. It basically reloads the data like a restore from a backup.

keennay:
Thanks!

PR #358, opened 2022-08-31
gossip: collect metric for block age

hadv:


hkalina:
@hadv Hi, are you sure you want this to be a `Timer`? To measure percentile/median values from a bunch of last processed blocks?
I would think it should be more a `Gauge`, to always show the last measured value - the age of the last processed block...

hadv:
> @hadv Hi, are you sure you want this to be a `Timer`? To measure percentile/median values from a bunch of last processed blocks? I would think it should be more a `Gauge`, to always show the last measured value - the age of the last processed block...

yeah, you're right. I update code to use gauge then we can show the value of block age instead of using percentile/median

hkalina (approved):

rus-alex (approved):

PR #363, opened 2022-09-06
Chain validation metric

rus-alex:
Here is the set of geth compatible metrics:
 - there are no uncle blocks in Opera, so "chain/reorg/*" are always zero;
 - "chain/validation" shows event validation time;
 
( it is an adaptation of #290 after #358 )

### DIFF
--- a/gossip/c_block_callbacks.go
+++ b/gossip/c_block_callbacks.go
@@ -53,6 +53,11 @@ var (
 	blockExecutionTimer = metrics.GetOrRegisterTimer("chain/execution", nil)
 	blockWriteTimer     = metrics.GetOrRegisterTimer("chain/write", nil)
 	blockAgeGauge       = metrics.GetOrRegisterGauge("chain/block/age", nil)
+
+	_ = metrics.GetOrRegisterMeter("chain/reorg/executes", nil)
+	_ = metrics.GetOrRegisterMeter("chain/reorg/add", nil)
+	_ = metrics.GetOrRegisterMeter("chain/reorg/drop", nil)
+	_ = metrics.GetOrRegisterMeter("chain/reorg/invalidTx", nil)
 )
 
 type ExtendedTxPosition struct {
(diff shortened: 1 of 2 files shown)


gossip/c_event_callbacks.go
```diff
@@ -80,18 +87,26 @@ func (s *Service) processSavedEvent(e *inter.EventPayload, es *iblockproc.EpochS
 		return errWrongMedianTime
 	}
 
+	begin := ctx.Value("startOfValidation").(time.Time)
+	blockValidationTimer.Update(time.Since(begin))
+
 	// aBFT processing
-	return s.engine.Process(e)
+	err = s.engine.Process(e)
```
hadv, line 94: Can you explain why we need to do this refactoring?

rus-alex: We don't need it )

hadv (approved):
LGTM

gossip/c_block_callbacks.go
```diff
@@ -53,6 +53,11 @@ var (
 	blockExecutionTimer = metrics.GetOrRegisterTimer("chain/execution", nil)
 	blockWriteTimer     = metrics.GetOrRegisterTimer("chain/write", nil)
 	blockAgeGauge       = metrics.GetOrRegisterGauge("chain/block/age", nil)
+
+	_ = metrics.GetOrRegisterMeter("chain/reorg/executes", nil)
+	_ = metrics.GetOrRegisterMeter("chain/reorg/add", nil)
+	_ = metrics.GetOrRegisterMeter("chain/reorg/drop", nil)
+	_ = metrics.GetOrRegisterMeter("chain/reorg/invalidTx", nil)
```
hkalina, line 60: @rus-alex what is the point in adding metrics, which will be unused?

hkalina: I agree we should add as much metrics from geth as we can, however if we cannot fill the metric with any data and we create it unused, it seems useless to me - is there some usecase which needs them to be present, even through they will have no meaningful value?

rus-alex: It just method to say: we didn't forget implemented such metrics and they better than geth's.
Agreed, it is overmach.

PR #366, opened 2022-09-19
Add parameter for setting time limit for RPC execution

hkalina:
Allow to configure timeout for RPC method calls.

Related to https://github.com/Fantom-foundation/go-ethereum/pull/38

(go.mod to be rebased after the related go-ethereum change is merged/tagged)


hkalina:
@uprendis is this acceptable?

PR #382, opened 2022-10-28
Fix: Invalid unit name while exporting Genesis

Dracoonghost:
While exporting genesis the unit names get Duplicated for the EVM section that causes an "Duplicate unit name" error while importing.


uprendis:
Hello @Dracoonghost,
Good day to you!
This bug appears to be fixed since 1.1.2-rc.1, although not released yet

PR #384, opened 2022-11-02
Gradual compaction

uprendis:
- Separate compaction into 256 prefixes, where each prefix is separated into 32 iterations, depending on range between first and last keys of the prefix. Previously compaction was separated only in 256 prefixes, which resulted into overly large compaction ranges, because data distribution isn't uniform and too much data may be stored behind the same prefix
- Add forced DB compaction after genesis processing and DB transformation/migration, because Pebble DB has performance issues after inserting a lot of data without manual compaction


hadv (approved):
SGTM

utils/compactdb/compactdb.go
```diff
@@ -0,0 +1,105 @@
...
+			keySize = len(first)
+		}
+		first = common.RightPadBytes(first, keySize-len(first))
+		last = common.RightPadBytes(last, keySize-len(last))
+		firstBn := new(big.Int).SetBytes(first)
+		lastBn := new(big.Int).SetBytes(last)
+		diff := new(big.Int).Sub(lastBn, firstBn)
+		if diff.Cmp(big.NewInt(10000)) < 0 {
+			// short circuit if too few keys
+			err := compact(b, nil, nil)
```
hadv, line 88: I think this might not work as expectation. It will compact the whole db, not only the current prefix, right?

uprendis: You're right

hadv (approved):

PR #400, opened 2022-12-09
Cache cap

uprendis:
- `--cache flag` is capped to 60% of available memory in order to prevent OEM. Previously `--cache` flag was meant to reflect upper bound of memory consumption but since 1.1.2, due to general increasing of DB caches, it's closer to average consumption (but still above it). Also, setting too high cache value may lead to worse performance because it limits FS cache
- print warning if `--cache` flag is absent and available memory allows more cache than default value

### DIFF
--- a/cmd/opera/launcher/config.go
+++ b/cmd/opera/launcher/config.go
@@ -411,7 +411,7 @@ func cacheScaler(ctx *cli.Context) cachescale.Func {
 		log.Crit("Invalid flag", "flag", CacheFlag.Name, "err", fmt.Sprintf("minimum cache size is %d MB", baseSize))
 	}
 	if totalMemory != 0 && targetCache > maxCache {
-		log.Warn(fmt.Sprintf("Requested cache size exceeds 60%% of availible memory. Reducing cache size to %d MB.", maxCache))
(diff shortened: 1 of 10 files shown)


cmd/opera/launcher/config.go
```diff
@@ -391,14 +393,28 @@ func nodeConfigWithFlags(ctx *cli.Context, cfg node.Config) node.Config {
...
+		}
 		return cachescale.Identity
 	}
-	targetCache := ctx.GlobalInt(CacheFlag.Name)
-	baseSize := DefaultCacheSize
 	if targetCache < baseSize {
 		log.Crit("Invalid flag", "flag", CacheFlag.Name, "err", fmt.Sprintf("minimum cache size is %d MB", baseSize))
 	}
+	if totalMemory != 0 && targetCache > maxCache {
+		log.Warn(fmt.Sprintf("Requested cache size exceeds 60%% of availible memory. Reducing cache size to %d MB.", maxCache))
```
hadv: just a typo

```suggestion
		log.Warn(fmt.Sprintf("Requested cache size exceeds 60% of available memory. Reducing cache size to %d MB.", maxCache))
```

hadv (approved):

PR #405, opened 2022-12-12
Refactoring: statedb interface

rus-alex:
Replaces exact `state.StateDB` struct with more general `evmcore.StateDB` interface.
This is necessary for easier maintain go-opera on different state db implementation (flat, erigon, etc.).

### DIFF
--- a/cmd/opera/launcher/export.go
+++ b/cmd/opera/launcher/export.go
@@ -2,6 +2,7 @@ package launcher
 
 import (
 	"compress/gzip"
+	"errors"
 	"io"
 	"os"
 	"strconv"
@@ -140,7 +141,7 @@ func exportEvmKeys(ctx *cli.Context) error {
 
 	evmDb := gdb.EvmStore().EvmDb
 	if evmDb == nil {
-		panic("unknown EvmDb format")
+		return errors.New("only legacy MPT EVM keys export makes sense")
 	}
 	it := evmDb.NewIterator(nil, nil)
 	// iterate only over MPT data
--- a/cmd/opera/launcher/export.go
+++ b/cmd/opera/launcher/export.go
@@ -138,7 +138,11 @@ func exportEvmKeys(ctx *cli.Context) error {
 	keysDB := batched.Wrap(keysDB_)
 	defer keysDB.Close()
 
-	it := gdb.EvmStore().EvmDb.NewIterator(nil, nil)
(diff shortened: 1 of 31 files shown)


cmd/opera/launcher/export.go
```diff
@@ -138,7 +138,11 @@ func exportEvmKeys(ctx *cli.Context) error {
 	keysDB := batched.Wrap(keysDB_)
 	defer keysDB.Close()
 
-	it := gdb.EvmStore().EvmDb.NewIterator(nil, nil)
+	evmDb := gdb.EvmStore().EvmDb
+	if evmDb == nil {
+		panic("unknown EvmDb format")
+	}
```
hadv: We might wrap these into a function in `evmstore.Store` to avoid duplicate code?

hadv (changes requested):

cyberbono3 (approved):

cmd/opera/launcher/export.go
rus-alex: Better I will add different error messages which explain each case.

hadv (approved):

ethapi/api.go
```diff
@@ -710,8 +709,14 @@ type StorageResult struct {
...
 // GetProof returns the Merkle-proof for a given account and optionally some storage keys.
 func (s *PublicBlockChainAPI) GetProof(ctx context.Context, address common.Address, storageKeys []string, blockNrOrHash rpc.BlockNumberOrHash) (*AccountResult, error) {
-	state, _, err := s.b.StateAndHeaderByNumberOrHash(ctx, blockNrOrHash)
-	if state == nil || err != nil {
+	statedb, _, err := s.b.StateAndHeaderByNumberOrHash(ctx, blockNrOrHash)
+	if statedb == nil || err != nil {
+		return nil, err
+	}
+
+	state, ok := evmcore.IsMptStateDB(statedb)
```
JekaMas, line 717: Sorry for asking, was just looking into the code and the consensus and found PRs.
How do you think, does it worth it to use generics+interfaces here to get rid of type assertion and have more strict type checks?

rus-alex: It will be an over-engineering for now. All we need is check here does the proofs complies with the legacy format.
We should not return others kind of proofs because clients understand the legacy (MPT) only.

PR #419, opened 2023-01-19
fix invalid character dash in prometheus metric names

### LINKED ISSUE #415: Invalid character dash in Prometheus metric names
**Describe the bug**
There are metrics in the Prometheus endpoint with the invalid character dash. Example: `opera_chaindata_gossip-17275_disk_size`.
Prometheus interface will complain with `invalid metric type "17275_disk_size gauge"`.

**To Reproduce**
Run txtracing binary from https://github.com/Fantom-foundation/go-opera/tree/release/txtracing/1.1.1-rc.2
Enable metrics and use Prometheus to scrape it at path `/debug/metrics/prometheus`.

**Expected behavior**
Prometheus able to scrape the metrics.

hadv:
to fix #415

### DIFF
--- a/integration/metric.go
+++ b/integration/metric.go
@@ -148,7 +148,7 @@ func (db *DBProducerWithMetrics) OpenDB(name string) (kvdb.Store, error) {
 	dm := WrapStoreWithMetrics(ds)
 	// disk size gauge should be meter separatly for each db name; otherwise,
 	// the last db siae metric will overwrite all the previoius one
-	dm.diskSizeGauge = metrics.GetOrRegisterGauge("opera/chaindata/"+name+"/disk/size", nil)
+	dm.diskSizeGauge = metrics.GetOrRegisterGauge("opera/chaindata/"+strings.ReplaceAll(name, "-", "_")+"/disk/size", nil)
 	if strings.HasPrefix(name, "gossip-") || strings.HasPrefix(name, "lachesis-") {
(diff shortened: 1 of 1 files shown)


rus-alex (approved):
@hadv , maybe it would be better to solve the problem basically?
The ways I see:
 - extend [mutateKey() func](https://github.com/Fantom-foundation/go-ethereum/blob/2b54480a4cff9997cb4377cb02abcffc8da8a597/metrics/prometheus/collector.go#L118) to replace all illegal chars (and PR to the ethereum/go-ethereum also); 
 - make go-opera uses custom metrics registry (instead of go-ethereum/metrics.DefaultRegistry) which will check names on register stage;

hadv:
> @hadv , maybe it would be better to solve the problem basically? The ways I see:
> 
> * extend [mutateKey() func](https://github.com/Fantom-foundation/go-ethereum/blob/2b54480a4cff9997cb4377cb02abcffc8da8a597/metrics/prometheus/collector.go#L118) to replace all illegal chars (and PR to the ethereum/go-ethereum also);
> * make go-opera uses custom metrics registry (instead of go-ethereum/metrics.DefaultRegistry) which will check names on register stage;

Yes, I sound good but basically we should choose properly metrics name for grafana/prometheus. 
So for simply, I think we just keep it simple like current fix.

uprendis:
Duplicates #418

PR #425, opened 2023-02-10
Global DB threads pool

rus-alex:
It is more general version of hotfix #414 :
- `kvdb/threads` db wrapper counts opened db iterators by global pool;
- whole go-opera db is wrapped with `kvdb/threads`, but it doesn't affect legacy: warn only if global pool exhausted;
- `topicsdb` search reserves the required number of iterators before start to work. Waits or fails if global pool exhausted;


kvdb/threads/pool_test.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+func TestThreadPool(t *testing.T) {
+	require := require.New(t)
+
+	for name, pool := range map[string]ThreadPool{
+		"global": GlobalPool,
+		"local":  ThreadPool{},
+	} {
+		t.Run(name, func(t *testing.T) {
```
hadv: @rus-alex I try to run but this test seems not running then it always passed with all inputs. I try to set [the line](https://github.com/Fantom-foundation/go-opera/pull/425/files#diff-1fd326c8be2df3ddbbbb287834a77dcc0e3a1a33dfe7ecba4e2620cb7fafbd9bR22)  like this `require.Equal(pool.Cap(), 100)` but still no test failed.

And if we set the max threads to 10 then the pool cap should be 8, right?

rus-alex: @hadv , thanks a lot, I forgot enable testing in TestMain()

hadv (approved):

cyberbono3 (approved):
LGTM

PR #426, opened 2023-02-15
fix of db_size metric calc

rus-alex:
Moves db_size metric calculation into singleton infinite loop, not on demand.
It is necessary because mainnet data dir is enough large and walking over the file tree takes a time.
Fixes the issue when metrics endpoint is really slow to respond.

### DIFF
--- a/cmd/opera/launcher/metrics/metrics.go
+++ b/cmd/opera/launcher/metrics/metrics.go
@@ -4,27 +4,40 @@ import (
 	"os"
 	"path/filepath"
 	"sync/atomic"
+	"time"
 
 	"github.com/ethereum/go-ethereum/log"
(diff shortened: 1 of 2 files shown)


hadv (approved):
SGTM

Maharacha:
I have the problem when syncing an archive node, either using the 3T genesis, or using a 7T snapshot, the node CPU usage is at 100% and the memory increases until OOM. If I turn off the metrics by removing the metrics CLI arguments everything works fine. Would this fix solve this problem?

Maharacha:
I just tried it and it works! Wow, so many hours I have spent on troubleshooting this. Big thanks!

alokeutpal (approved):

PR #428, opened 2023-02-22
fix non-determistic fakenet validator key with Go 1.20

hadv:
There's a change of package crypto/ecdsa from Go 1.20.1 that not generating deterministic keys using GenerateKey(). On the fakenet, we call get fake validator keys in 3 difference places (genesis, config and keystore) so the fake validator keys were difference, which make the error.

We need to find a way to generate deterministic keys or generating only one time and re-use in other place. The issue might happen again when user running fakenet on multiple validators then the fake validators on each fakenet node might be difference as well.

So we might to hard code the fakenet validator keys, but the issue is the total number of validators is not fixed but for testing I think we can limit to maximum 10 validators?


hadv:
**_Note:_** by the way, some tests on https://github.com/Fantom-foundation/go-opera/blob/develop/gossip/mps_test.go and https://github.com/Fantom-foundation/go-opera/blob/master/gossip/heavycheck_test.go will be failed if we use random fakenet validators. It seems that we has some specific assumption on the test code

But if we generate deterministic validator keys from old version of `crypto` lib, then all the test passed now.

evmcore/apply_fake_genesis.go
```diff
@@ -34,6 +34,121 @@ import (
...
+	key30, key31, key32, key33, key34, key35, key36, key37, key38, key39,
+	key40, key41, key42, key43, key44, key45, key46, key47, key48, key49,
+	key50, key51, key52, key53, key54, key55, key56, key57, key58, key59,
+	key60, key61, key62, key63, key64, key65, key66, key67, key68, key69,
+	key70, key71, key72, key73, key74, key75, key76, key77, key78, key79,
+	key80, key81, key82, key83, key84, key85, key86, key87, key88, key89,
+	key90, key91, key92, key93, key94, key95, key96, key97, key98, key99,
+	key00,
+}
+
```
cyberbono3: I would put variables `keys` and `key00..key99` inside `FakeKey` to make them local as they are accessed only  in `FakeKey`.  Why you add `key00` second time in L149?

hadv: Okay, `key00` actually is `key100`, the `key00` at zero index is not used because the validator key start from 1

evmcore/apply_fake_genesis.go
```diff
@@ -93,12 +94,123 @@ func MustApplyFakeGenesis(statedb *state.StateDB, time inter.Timestamp, balances
...
+		key10, key11, key12, key13, key14, key15, key16, key17, key18, key19,
+		key20, key21, key22, key23, key24, key25, key26, key27, key28, key29,
+		key30, key31, key32, key33, key34, key35, key36, key37, key38, key39,
+		key40, key41, key42, key43, key44, key45, key46, key47, key48, key49,
+		key50, key51, key52, key53, key54, key55, key56, key57, key58, key59,
+		key60, key61, key62, key63, key64, key65, key66, key67, key68, key69,
+		key70, key71, key72, key73, key74, key75, key76, key77, key78, key79,
+		key80, key81, key82, key83, key84, key85, key86, key87, key88, key89,
+		key90, key91, key92, key93, key94, key95, key96, key97, key98, key99,
+	}
```
cyberbono3: I would change the type of `n` to `uint` to satisfy your condition at L211 as `n` can not be negative. I think, there is an idiomatic way to declare multiple variables inside `FakeKey` like that:
```
var (
     key = ...
     keys = [100]*ecdsa.PrivateKey{
)
```

hadv: > I think, there is an idiomatic way to declare multiple variables inside FakeKey like that:

Okay, I write code to generate this code segment so didn't notice this

> I would change the type of n to uint 

`idx.ValidatorID` type actually is `uint32` to I changed the type of `n` to `uint32` as well

cyberbono3 (approved):
SGTM

evmcore/apply_fake_genesis.go
quan8: Wondering if these will work:

var seed = []byte("163f5f0f9a621d72fedd85ffca3d08d131ab4e812181e0d30ffd1c885d20aac73144c0aa4ced56dc15c79b045bc5559a5ac9");

.......

// FakeKey gets n-th fake private key.
func FakeKey(n uint32) *ecdsa.PrivateKey {
	if n == 0 || n > 100 {
			panic(errors.New("validator num is out of range"))
	}

	// shift left n characters of the seed to the end
	var shifted = string(append(seed[n:], seed[:n]...))
	// and get the first 64 bytes
	var k = "0x" + string(shifted[:64]);

	var key, err = crypto.ToECDSA(hexutil.MustDecode(k))

 	if err != nil {
 		panic(err)
 	}

	return key
}

hadv: Yes, I think it might work but we want to keep the **fakekey** as same as when running below code on old version of golang. So we pre-generated all the first 100 keys from old version of golang and hardcode into the source like that.

```
	reader := rand.New(rand.NewSource(int64(n)))

	key, err := ecdsa.GenerateKey(crypto.S256(), reader)
```

PR #431, opened 2023-02-27
Disconnect invalid and inactive peers

jmozah:
This PR adds checks to identify and ban peers that pass the P2P handshake and are accepted into the application protocol but has other application-level issues.

- Some clients have valid caps (opera/63, fsnap/1) but invalid client names such as Efireal, go-corex, Geth etc.
- Progress message is checked if the Epoch increases for a nominal duration.
- Application message should be received within a threshold.
- Recurring Application error now results in banning the peer.

These checks have shown that peers that are valid and working honestly get priority.

Depends on https://github.com/Fantom-foundation/go-ethereum/pull/44

### DIFF
--- a/gossip/handler.go
+++ b/gossip/handler.go
@@ -63,14 +63,14 @@ const (
 	txChanSize = 4096
 
 	// percentage of useless peer nodes to allow
-	uselessPeerPercentage = 20 // 20%
+	uselessPeerPercentage = 0.2 // 20%
 
 	// Number of application errors that can be tolerated before banning the node and disconnecting
 	toleranceOfApplicationErrors = 3
 )
 
 var (
-	ErrorProgressTimeout = errors.New("progress timeout")
+	ErrorProgressTimeout    = errors.New("progress timeout")
 	ErrorApplicationTimeout = errors.New("application timeout")
 )
 
@@ -795,7 +795,7 @@ func (h *handler) handle(p *peer) error {
 	}
 
 	// Some clients have compatible caps and thus pass discovery checks and seep in to
-	// protocol handler. We should band these clients immediately.
+	// protocol handler. We should ban these clients immediately.
 	// ex: go-corex, Efireal, Geth all with caps=[opera/62]
 	if !strings.Contains(strings.ToLower(p.Name()), "opera") {
 		discfilter.Ban(p.ID())
@@ -804,7 +804,7 @@ func (h *handler) handle(p *peer) error {
 
 	// A useless peer is the one which does not support protocols opera/63 & fsnap/1.
 	useless := !eligibleForSnap(p.Peer)
-	if !p.Peer.Info().Network.Trusted && useless && h.peers.UselessNum() >= (h.maxPeers*(uselessPeerPercentage/100)) {
+	if !p.Peer.Info().Network.Trusted && useless && h.peers.UselessNum() >= int((float32(h.maxPeers)*uselessPeerPercentage)) {
 		// don't allow more than 20% of useless peers
 		return p2p.DiscTooManyPeers
 	}
@@ -875,10 +875,10 @@ func (h *handler) handle(p *peer) error {
 	// Handle incoming messages until the connection is torn down or the inactivity
 	// timer times out.
 	var noOfApplicationErrors = 0
+	// progress and application
+	progressWatchDogTimer := time.NewTimer(noProgressTime)
+	applicationWatchDogTimer := time.NewTimer(noAppMessageTime)
 	for {
-		// progress and application
-		progressWatchDogTimer := time.NewTimer(noProgressTime)
-		applicationWatchDogTimer := time.NewTimer(noAppMessageTime)
 		select {
 		case <-progressWatchDogTimer.C:
 			// If self syncing, don't check peer progress
@@ -893,7 +893,7 @@ func (h *handler) handle(p *peer) error {
 				discfilter.Ban(p.ID())
 				return ErrorProgressTimeout
 			}
-		case <- applicationWatchDogTimer.C:
+		case <-applicationWatchDogTimer.C:
 			if p.IsApplicationProgressing() {
 				applicationWatchDogTimer.Reset(noAppMessageTime)
 			} else {
@@ -1072,7 +1072,7 @@ func (h *handler) handleMsg(p *peer) error {
 		p.SetProgress(progress)
 		// If peer has not progressed for noProgressTime minutes, then disconnect the peer.
 		if !p.IsPeerProgressing() {
-			return errResp(ErrPeerNotProgressing, "%v: %v %v", "epoch is not progressing for ", noProgressTime, "minutes")
+			return errResp(ErrPeerNotProgressing, "%v: %v", "epoch is not progressing for ", noProgressTime)
 		}
 
 	case msg.Code == EvmTxsMsg:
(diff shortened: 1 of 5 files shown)


gossip/handler.go
```diff
@@ -782,23 +793,30 @@ func (h *handler) handle(p *peer) error {
 		p.Log().Error("Snapshot extension barrier failed", "err", err)
 		return err
 	}
-	useless := discfilter.Banned(p.Node().ID(), p.Node().Record())
-	if !useless && (!eligibleForSnap(p.Peer) || !strings.Contains(strings.ToLower(p.Name()), "opera")) {
-		useless = true
+
+	// Some clients have compatible caps and thus pass discovery checks and seep in to
+	// protocol handler. We should band these clients immediately.
```
holisticode: nit: little typo

gossip/handler.go
```diff
@@ -61,6 +61,17 @@ const (
 	// txChanSize is the size of channel listening to NewTxsNotify.
 	// The number is referenced from the size of tx pool.
 	txChanSize = 4096
+
+	// percentage of useless peer nodes to allow
+	uselessPeerPercentage = 20 // 20%
```
holisticode: Nit: Why don't we use just a factor, e.g. 0.2, instead of then having to calculate each time the percentage?

gossip/handler.go
```diff
@@ -782,23 +793,30 @@ func (h *handler) handle(p *peer) error {
...
+	if !strings.Contains(strings.ToLower(p.Name()), "opera") {
 		discfilter.Ban(p.ID())
+		return p2p.DiscProtocolError
 	}
-	if !p.Peer.Info().Network.Trusted && useless && h.peers.UselessNum() >= h.maxPeers/10 {
-		// don't allow more than 10% of useless peers
+
+	// A useless peer is the one which does not support protocols opera/63 & fsnap/1.
+	useless := !eligibleForSnap(p.Peer)
+	if !p.Peer.Info().Network.Trusted && useless && h.peers.UselessNum() >= (h.maxPeers*(uselessPeerPercentage/100)) {
```
holisticode: Question: I am not yet familiar with this useless stuff, but why do we even allow a percentage of useless peers at all? Why don't we just disconnect them all?

gossip/handler.go
```diff
@@ -856,11 +872,51 @@ func (h *handler) handle(p *peer) error {
...
+	// Handle incoming messages until the connection is torn down or the inactivity
+	// timer times out.
+	var noOfApplicationErrors = 0
 	for {
-		if err := h.handleMsg(p); err != nil {
-			p.Log().Debug("Message handling failed", "err", err)
-			return err
+		// progress and application
+		progressWatchDogTimer := time.NewTimer(noProgressTime)
+		applicationWatchDogTimer := time.NewTimer(noAppMessageTime)
```
holisticode: Aren't we recreating the timer on each `for` iteration here? Therefore the `Reset`s later are useless? It looks to me that either we have to create the timers outside of the for loop, and then `Reset` them as you do now, or recreating them in each loop iteration and just break when we `Reset`, although this then results in a lot of garbage collected timers? Or am I missing something?

gossip/handler.go
```diff
@@ -856,11 +872,51 @@ func (h *handler) handle(p *peer) error {
...
+			} else {
+				p.Log().Warn("application timer timeout: ", "name", p.Name(), "node", p.Node().String())
+				discfilter.Ban(p.ID())
+				return ErrorApplicationTimeout
+			}
+		default:
+			err := h.handleMsg(p)
+			if err != nil {
+				p.Log().Debug("Message handling failed", "err", err)
+				if strings.Contains(err.Error(), errorToString[ErrPeerNotProgressing]) {
```
holisticode, line 908: Can we use `errors.Is` here instead of comparing strings?

gossip/handler.go
```diff
@@ -1014,6 +1070,10 @@ func (h *handler) handleMsg(p *peer) error {
 			return errResp(ErrDecode, "%v: %v", msg, err)
 		}
 		p.SetProgress(progress)
+		// If peer has not progressed for noProgressTime minutes, then disconnect the peer.
+		if !p.IsPeerProgressing() {
+			return errResp(ErrPeerNotProgressing, "%v: %v %v", "epoch is not progressing for ", noProgressTime, "minutes")
```
holisticode: Nit: As `noProgressTime` is a duration, this would print "epoch is not progressing for 3m0s minutes", I think

gossip/handler.go
```diff
@@ -1316,6 +1376,11 @@ func (h *handler) handleMsg(p *peer) error {
 	default:
 		return errResp(ErrInvalidMsgCode, "%v", msg.Code)
 	}
+
+	if msg.Code != ProgressMsg {
```
holisticode, line 1380: I am not yet familiar with all message codes, but is `ProgressMsg` the only message which signals that there is progress?

gossip/peer.go
```diff
@@ -85,7 +92,29 @@ func (p *peer) SetProgress(x PeerProgress) {
...
 
+	// Check if the peer is progressing
+	if x.More(p.progress) {
+		p.setPeerAsProgressing(x)
+	}
+}
+
+func (p *peer) setPeerAsProgressing(x PeerProgress) {
 	p.progress = x
+	p.progressTime = time.Now()
```
holisticode, line 103: Any specific reason why `p.appMessageTime` is locked, but `p.progressTime` isn't?

gossip/peer_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestPeerProgressWithEpoch(t *testing.T) {
+	// Increment epoch and see if the peer is progressing
+	setProgressThreshold(1 * time.Second)
+	newPeer := getPeer()
+	ep1 := PeerProgress{Epoch: 1}
+	newPeer.SetProgress(ep1)
+	time.Sleep(2 * time.Second) //set the threshold to 2 second
```
holisticode: All these `Sleep` acctumulate to 9 seconds - making test runs 9 seconds slower as I understand. Isn't there a different way to test this? Do we actually even need to sleep?

gossip/handler.go
jmozah: Well, the peer is useless in the context of sync, i.e. it doesn't support `fsnap/1` and `opera/63`.
But old peers supporting `opera/62` should still be allowed to participate.

gossip/handler.go
jmozah: Oops... the timer should be outside the loop.

gossip/handler.go
jmozah: We can use `errors.Is()` only to compare errors. But in this place, the error is defined as a string.
If we want to change it, we should define all the errors as errors.New().

gossip/handler.go
jmozah: Yes

gossip/peer.go
jmozah: It's locked in `SetProgress()` where `setPeerAsProgressing()` is called.

gossip/handler.go
holisticode: Ah so I assume `useless` then already checked that the peer is a `opera/62` peer. It's not just any peer. That would make sense.

gossip/handler.go
holisticode: Yes agreed. If there are more such string based errors instead of `errors.New()` based ones (which I believe would be better) - then this should go into a separate PR to address. So up to you if you want to do anything in this PR.

holisticode (approved):
I am not sure if I should already be the only person approving, but I want to signal that this looks good to me now (at least).

PR #436, opened 2023-03-08
current version only can be build with go1.19+

hadv:


hadv:
better fix to compatibility with g1.17 on #434

PR #462, opened 2023-05-30
Changes to allow extra testing

holisticode:
This PR introduces some changes, which are required to allow extra testing.

The testing peer can send events to test other peers. The testing peer is running as a full node.

This PR allows to access some internals of the peer, and it changes the demo network (declares 4 validators -- starting with 3 and add additional one).


rus-alex (changes requested):
@holisticode , I got your intent but suggest some style changes before merge: https://github.com/holisticode/go-opera/pull/4

Issue #461, opened 2023-05-27
Unique metric streams are unbounded


thingtheory:
**Describe the bug**

`opera` creates multiple metric streams (in this case, unique metrics) per epoch. This means that a continuous running node will create, and report to consumers, an ever increasing number of streams (until it restarts).

Example (prometheus style)

```
# TYPE opera_chaindata_epoch-213180_disk_read gauge
opera_chaindata_epoch-213180_disk_read 7820353

# TYPE opera_chaindata_epoch-213180_disk_write gauge
opera_chaindata_epoch-213180_disk_write 35209142

# TYPE opera_chaindata_epoch_213180_disk_size gauge
opera_chaindata_epoch_213180_disk_size 3982687
```

**To Reproduce**

Run `opera` with `--metrics` to enable metrics.

**Expected behavior**

The number of streams reported by `opera` does not increase over time.

There's probably a few ways to solve this:

1) Don't report per epoch metrics
2) Put the per epoch metrics behind the `--metrics.expensive` or new flag
3) Only report metrics for the current or last N epochs

**Additional info**

I'm not familiar with how influxdb works, but this behavior causes prometheus memory usage to grow unbounded due to the increasing number of *active* streams.

PR #465, opened 2023-06-08
Fix of tmp db size metric

rus-alex:
Replaces infinite number of metrics for each tmp DB with single generic metric for each kind of tmp DB.
Makes it easier to build a dashboard,  prevents the leaks like #461


integration/metric.go
```diff
@@ -140,22 +141,34 @@ func (ds *StoreWithMetrics) meter(refresh time.Duration) {
 	errc <- merr
 }
 
+var tmpDbNameMask = regexp.MustCompile("^([A-z]+)(-[0-9]+)$")
+
+func genericNameOfTmpDB(name string) string {
+	match := tmpDbNameMask.FindStringSubmatch(name)
+	if len(match) == 3 {
+		return match[1] + "es"
```
hadv: Ending with "es" looks a bit strange for `gossipes` and `epoches`.

We might include all (gossip-, lachesis- and  epoch-) into `epochs` like before?

rus-alex: How about gossip-tmp, lachesis-tmp and epoch-tmp ?

thaarok: @rus-alex I would say it is more like "gossip-epoch", "lachesis-epoch" and "epoch".
But I like the idea of one "epochs" generic name more.

rus-alex: @hkalina , but `gossip-%d` and `lachesis-%d` are not the same DB for "legacy-ldb" and "legacy-pbl" layouts. So we should have different metrics for each, not a common "epochs".
I agree that we mostly use "ldb-1" and "pbl-1" now. But it may be changed later.

rus-alex: Or, we may add a new kind of temporary db.

thaarok (approved):

PR #469, opened 2023-06-14
Update redundant ports in Dockerfile

holisticode:
It looks like the file `docker/Dockefile.opera`  is obsolete. This PR proposes to remove it.


rus-alex:
Why do you think it is obsolete?
Why did you choose to remove it rather than update?
If remove then remove `make opera-image` command also.

rus-alex (approved):

PR #496, opened 2023-08-08
Fix removing from txpool priced structure and localGauge

thaarok:
Few fixes of txpool:
* `localGauge` should be increased when remote txs are migrated into locals
* `priced` should get the removed notification even for removing in `runReorg()/demoteUnexecutables()` - it is not necessary when the MinGasPrice changes (as the change trigger priced reheaping), but when the MinGasPrice stays unchanged for a long time, it leads to memory exhaustion by the priced structure grow.


thaarok:
Hi @uprendis, please consider merging this fixes. Any feedback welcome. Thanks!

thaarok:
Fixed: the removeTx boolean parameter was correct, reverted.

PR #500, opened 2023-08-11
Remove redundant DefaultDataDir call

holisticode:
The `DefaultDataDir()` call is already assigned in the `NodeDefaultConfig` global variable on line 32 in `defaults.go`

### DIFF
--- a/cmd/opera/launcher/config.go
+++ b/cmd/opera/launcher/config.go
@@ -563,7 +563,6 @@ func defaultNodeConfig() node.Config {
 	cfg.HTTPModules = append(cfg.HTTPModules, "eth", "ftm", "dag", "abft", "web3")
 	cfg.WSModules = append(cfg.WSModules, "eth", "ftm", "dag", "abft", "web3")
 	cfg.IPCPath = "opera.ipc"
-	cfg.DataDir = DefaultDataDir()
 	return cfg
 }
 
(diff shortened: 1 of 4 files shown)


cmd/opera/launcher/config.go
```diff
@@ -563,7 +563,6 @@ func defaultNodeConfig() node.Config {
 	cfg.HTTPModules = append(cfg.HTTPModules, "eth", "ftm", "dag", "abft", "web3")
 	cfg.WSModules = append(cfg.WSModules, "eth", "ftm", "dag", "abft", "web3")
 	cfg.IPCPath = "opera.ipc"
-	cfg.DataDir = DefaultDataDir()
```
rus-alex, line 566: Nice change.

rus-alex (approved):
Why don't you refactor this part totally along with it?
Existence of `defaultNodeConfig()` and `NodeDefaultConfig` at the same time is confusing. Let's replace them with single `DefaultNodeConfig()` for example.

holisticode:
@rus-alex refactored according to your suggestion

cmd/opera/launcher/defaults.go
```diff
@@ -28,22 +29,27 @@ func overrideFlags() {
 }
 
 // NodeDefaultConfig contains reasonable default settings.
```
rus-alex: Godoc agreement: the comment should begins with the name of the element it describes.

holisticode: Updated and also fixed tests

PR #553, opened 2024-07-11
Patch for debug trace message to address reference

jenikd:
Fixing reference to message to address and adding recover function for tracetx function.
Also adding possibility for node to connect no Sonic nodes


b-scholz (approved):
LGTM

